PR #7, opened 2016-05-23
add cifar100-convnet example and its dataflow abstraction.

cykustcc:
* Add cifar100-convnet example.
* Add cifar100 in dataflow/dataset


ppwwyyxx:
Thanks a lot for sending the first pull request! 
However, examples are for illustration purpose of the usage of this library. Since the cifar100 example is almost identical to cifar10, it's not necessary to add an extra one specifically for cifar100.

Same for the dataflow code. We would really prefer minimal duplicated code and more modularity, because that's all this project is intended for. 
Something you may try:
Have `Cifar100` being a subclass of `Cifar10`, or more generally, use a  `CifarBase` and have a new `Cifar10` and `Cifar100` being subclasses of `CifarBase`. `Cifar10` and `Cifar100` would only be different in how they load the data, and that could be done by calling the corresponding function by different arguments (different URLs, different keys for the label, etc.) This way most existing code can be reused.

cykustcc:
Thanks! Now  ```CifarBase``` is the baseclass of ```Cifar10``` and ```Cifar100```. And the duplicate code in dataflow is also removed.

PR #22, opened 2016-09-20
Fix bug in get_output_tensors method of inference

### LINKED ISSUE #21: Bug in inference get_output_tensors method
1. _get_output_vars method is not defined, should be _get_output_tensors. 
2. InferenceRunner should call "vc.get_output_tensors()" as opposed to "vc._get_output_tensors()" in line 94.

sujaynarumanchi:
Fix #21


ppwwyyxx:
Fixed in fbf93d44c. Thanks!

PR #24, opened 2016-09-21
Minor bugfix imgaug

### LINKED ISSUE #23: Super class init method not called for some image augmentors
Several image augmentors (RandomCrop, RandomCropRandomShape ...) do not call super class init method, hence self.rng is not set until reset is explicitly called.
This makes usage in standalone mode infeasible with the error self.rng variable is not defined.

sujaynarumanchi:
Fix #23 

- [x] Add salt and pepper noise

### DIFF
--- a/tensorpack/dataflow/imgaug/noise.py
+++ b/tensorpack/dataflow/imgaug/noise.py
@@ -37,7 +37,8 @@ class GaussianNoise(ImageAugmentor):
         return ret
 
 class SaltPepperNoise(ImageAugmentor):
-    def __init__(self, amount=0.05, s_vs_p=0.5):
+    def __init__(self, white_prob=0.05, black_prob=0.05):
+        assert white_prob + black_prob <= 1, "Sum of probabilities cannot be greater than 1"
         super(SaltPepperNoise, self).__init__()
         self._init(locals())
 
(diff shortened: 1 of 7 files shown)


tensorpack/dataflow/imgaug/noise.py
```diff
@@ -33,3 +35,17 @@ def _augment(self, img, noise):
...
+class SaltPepperNoise(ImageAugmentor):
+    def __init__(self, amount=0.05, s_vs_p=0.5):
+        super(SaltPepperNoise, self).__init__()
+        self._init(locals())
+
+    def _get_augment_params(self, img):
+        return self.rng.uniform(low=0, high=1, size=img.shape)
+
+    def _augment(self, img, param):
+        ret = img.copy()
```
ppwwyyxx: The current contract is that augmentors are allowed to modify the image in-place to be efficient. In that case the lowest-level dataflow should avoid its data being damaged by augmentors (by yielding a copy, see `dataset/cifar.py`).

tensorpack/dataflow/imgaug/noise.py
```diff
@@ -33,3 +35,17 @@ def _augment(self, img, noise):
...
+        ret[param > (1 - self.amount * self.s_vs_p)] = 255
+        ret[param < (1 - self.s_vs_p) * self.amount] = 0
```
ppwwyyxx: Should it also be `(1 - self.amount * self.s_vs_p)`?

ppwwyyxx: Maybe it's easier to understand to use two argument `black_prob` and `white_prob` instead of `amount` and `s_vs_p`.

sujaynarumanchi: Agreed, have made the change in the latest commit.

tensorpack/dataflow/imgaug/noise.py
sujaynarumanchi: Sure, have made the changes.

PR #61, opened 2016-12-06
Lmdb Reader

PatWie:
Reading Lmdb databases in random order requires to shuffle the keys.
Note, "_txn.cursor()" returns (k,v) pairs. Currently, the
implementations tries to query by a (k,v) pair instead of the key
only.


ppwwyyxx:
Thanks for the fix!

PR #81, opened 2017-01-02
add tfSlim to examples

PatWie:
This handles non-gradient-descent updates like batch-norm from
tfSlim and regularization losses by fetching the op-collections.

### DIFF
--- a/tensorpack/models/model_desc.py
+++ b/tensorpack/models/model_desc.py
@@ -113,9 +113,44 @@ Use _build_graph(self, input_vars) and get_current_tower_context().is_training i
     def get_cost(self):
         """
         Return the cost tensor in the graph. Called by some of the :class:`tensorpack.train.Trainer` which
-        assumes single-cost models.
+        assumes single-cost models. Apply tfSlim modifications.
         """
-        return self._get_cost()
+
+        # current scope
+        scope = tf.get_variable_scope()
+
+        # the model cost so far
+        cost = self._get_cost()
+
+        # In contrast to this lib, when using tfSlim the user expect
+        #         "with slim.arg_scope([...], weights_regularizer=slim.l2_regularizer(0.001)"
+        # to regularize these layers automatically. Note, this already contains the multiplier!
+        regulization_losses = 0
+        # try to prevent regEx error, iff scope name is empty ("")
+        try:
+            regulization_losses = set(tf.get_collection(tf.GraphKeys.REGULARIZATION_LOSSES, scope=scope))
+        except Exception:
+            regulization_losses = set(tf.get_collection(tf.GraphKeys.REGULARIZATION_LOSSES))
+
+        # TODO: check if "scope=scope" should be used here too
+        if len(regulization_losses) > 0:
+            cost += tf.add_n(regulization_losses, name="regularize_loss")
+
+        # As these batch-norm statistics quickly accumulate, there is no significant loss of accuracy
+        # if only the main tower handles all batch-normalization updates, which are then shared across
+        # the towers
+        ctx = get_current_tower_context()
+        if ctx is not None and ctx.is_main_training_tower:
+            # if there is no entry in tf.GraphKeys.UPDATE_OPS, then there is a regEx exception
+            try:
+                non_grad_updates = set(tf.get_collection(tf.GraphKeys.UPDATE_OPS, scope=scope))
+            except Exception:
+                non_grad_updates = set(tf.get_collection(tf.GraphKeys.UPDATE_OPS))
+            if non_grad_updates:
+                    with tf.control_dependencies(non_grad_updates):
+                        barrier = tf.control_flow_ops.no_op(name='batchnorm_barrier')
+                    cost = tf.control_flow_ops.with_dependencies([barrier], cost)
+        return cost
 
     def _get_cost(self, *args):
         return self.cost
--- a/tensorpack/train/base.py
+++ b/tensorpack/train/base.py
@@ -155,6 +155,8 @@ class Trainer(object):
                     self.trigger_epoch()
             except StopTraining:
                 logger.info("Training was stopped.")
+            except KeyboardInterrupt:
+                logger.info("Detected Ctrl+C and shutdown training.")
             except:
                 raise
             finally:


PatWie:
see issue #84

examples/simple-mnist/mnist-convnet.py
```diff
@@ -0,0 +1,121 @@
...
+import argparse
+import tensorflow.contrib.slim as slim
+import os
+
+"""
+MNIST ConvNet example.
+using tfSlim for network definition with validation error: 0.5% after 25 epochs
+"""
+
+# Just import everything into current namespace
```
ppwwyyxx: We can remove this comment here because it's not `import *`.

PatWie:
Rebased on current HEAD after fixing #84

tensorpack/train/base.py
```diff
@@ -53,6 +55,21 @@ def __init__(self, config):
...
+        regul_loss = ops.get_collection(ops.GraphKeys.REGULARIZATION_LOSSES)
+        if len(regul_loss) > 0:
+            loss_op += tf.add_n(regul_loss)
+
+        # add non-gradient updates, like batch-norm
+        update_ops = set(ops.get_collection(ops.GraphKeys.UPDATE_OPS))
+        if update_ops:
+            with ops.control_dependencies(update_ops):
+                barrier = control_flow_ops.no_op(name='update_barrier')
+            loss_op = control_flow_ops.with_dependencies([barrier], loss_op)
```
ppwwyyxx: Is this equivalent to just `loss_op = control_flow_ops.with_dependencies(update_ops, loss_op)` ?

tensorpack/train/base.py
```diff
@@ -53,6 +55,21 @@ def __init__(self, config):
...
         self.coord = tf.train.Coordinator()
 
+    def apply_tfslim_collections(self, loss_op):
+        # add regularization loss
+        regul_loss = ops.get_collection(ops.GraphKeys.REGULARIZATION_LOSSES)
+        if len(regul_loss) > 0:
+            loss_op += tf.add_n(regul_loss)
+
+        # add non-gradient updates, like batch-norm
+        update_ops = set(ops.get_collection(ops.GraphKeys.UPDATE_OPS))
```
PatWie: @ppwwyyxx Are you worried about the scope? There is a scope-arg in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/g3doc/api_docs/python/functions_and_classes/shard7/tf.get_collection.md

tensorpack/train/base.py
PatWie: This is from
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/slim/python/slim/learning.py#L428-L430

I think it is just a nice way to name this barrier for TensorBoard.

PatWie:
About your points:

1. I put the batch_norm collection into the trainer because it has to be connected to the computation graph. I changed this in the last version of this pull-request.

2. You are right, this belongs to the model --> fixed. For the simple mnist example, I think it is sufficient to put it there. But in the long run, if you want to support tfSlim it might be good, to handle the regularization in the background. I mean, if `set(tf.get_collection(tf.GraphKeys.REGULARIZATION_LOSSES))` is empty, then there is nothing added to the graph. So this can be transparent.

3. Now, BN-updates are only considered from a single tower which is sufficient for updates. 

4. I am not sure if the updates version - a mixture of your layers and tfSlim - is easier to understand :wink:

examples/mnist-convnet.py
```diff
@@ -72,16 +86,23 @@ def _build_graph(self, input_vars):
...
+            wd_cost = tf.mul(1e-5,
+                             regularize_cost('fc.*/W', tf.nn.l2_loss),
+                             name='regularize_loss')
+            summary.add_moving_summary(cost, wd_cost)
+
+            # monitor histogram of all weight (of conv and fc layers) in tensorboard
+            summary.add_param_summary([('.*/W', ['histogram'])])
+            self.cost = tf.add_n([wd_cost, cost], name='cost')
+        else:
+            regul_loss = tf.get_collection(tf.GraphKeys.REGULARIZATION_LOSSES)
```
PatWie: I think it would be better to add this to any trainer,
In the model it can be turned off and on by

````python
with slim.arg_scope([slim.layers.conv2d, ...], normalizer_fn=slim.layers.batch_norm, weights_regularizer=slim.l2_regularizer(0.001),
````

File: tensorpack/train/multigpu.py [hunk 5]
```diff
@@ -44,8 +44,13 @@ def _multi_tower_grads(towers, get_tower_grad_func):
                 if idx == 0:
                     # avoid repeated summary from each device
                     backup = backup_collection(SUMMARY_BACKUP_KEYS)
+
+                # We could extract all batchnorm updates from all tower, but we just use one tower,
+                # which is sufficient
+                non_grad_update_op = ops.get_collection(ops.GraphKeys.UPDATE_OPS, scope=scope)
```
PatWie: Maybe this line can also be indented to be only on the first (main) tower

ppwwyyxx:
I think it's important to keep things controllable by the users -- at least the entire forward pass, instead of hiding it in the trainer.
For example, BN updates is not necessarily applied before the cost tensor: in my implementation it is applied inside BN layer. If UPDATE_OPS collection is used for some other purposes, differences like this may be important.
You can provide very strong utilities function such as `apply_tflim_collections`, or a subclass such as `SlimModelDesc` which automatically applies these in `get_cost`. But I'd still like the users to enable them explicitly.

PatWie:
I understand your point. But in tfSlin the only purpose of the update_ops collection is to manage batchnorm updates. In this sense the implementation from the pull request is totally equivalent to your BatchnormV2. As there is no way of overriding the slim layers without patching the tf-repo I have put it into the trainer.
Currently there is no overhead since it also only updates everything only in the main tower if the user asked for batchnorm.

Same for regularization: As long as there is no explizit regularization within in the ModelDesc, there are no aditional operations added to the graph.

I these two points I would argue that since it is well documented behavior in tfSlim one can add these in the backend. I would further say it is the only way to add these as the user expect these additional operations when using tfSlim.

If you still want, I can rebase again on your HEAD.

ppwwyyxx:
I see. Now I'm accepting the idea of applying the two collections by default, because:
1. Nothing different when the two collections are not used
2. We can assume these two collections are only used for these two purposes

But still, I hope the trainer keeps as simple as possible. An alternative way is to apply these collections in `ModelDesc.get_cost()` by default. This way it goes into models, and users will also have a way to disable it (by override the function). What do you think of this?

PatWie:
I put that new stuff into ModelDescr. However, this currently does not affect possible GAN-models. For the examples, I will probably create another repo not to mess up your code. You can then decide whether you want them or not.

I think ModelDesc should have something like GanModelDesc such that the cost-function can be overridden.

ppwwyyxx:
LGTM. I'll merge it soon. Thanks again! 
My idea is that the existing framework provides Trainer and ModelDesc for the most common use case (single cost optimization). For tasks more complicated than that I'll expect users to write some of their own code -- it'll just be too hard to write anything general.
For example, for GAN if you want to help users handle the collections, there is going to be a lot of issues about which cost should the collection be applied on -- ideally those appeared in generator should be on generator cost, and those appeared in discriminator should be on discriminator cost. But for extensions like mode-regularized GAN, there is an extra "encoder". For variants like Image2Image, BN statistics doesn't need update at all.  -- These are situations that cannot be simply addressed by an abstraction, so I'd rather just provide utility function and let users call them.

PatWie:
Please wait before merging, there are still some issues with the scope.

PatWie:
The last commit is tested with batch-norm and regularization:
https://github.com/PatWie/tensorpack-recipes/blob/master/mnist/run.py

tensorpack/models/model_desc.py
```diff
@@ -113,9 +113,37 @@ def _build_graph(self, inputs):
...
+
+        # As these batch-norm statistics quickly accumulate, there is no significant loss of accuracy
+        # if only the main tower handles all batch-normalization updates, which are then shared across
+        # the towers
+        scope = tf.get_variable_scope()
+        ctx = get_current_tower_context()
+        if ctx is not None and ctx.is_main_training_tower:
+            # if there is no entry in tf.GraphKeys.UPDATE_OPS, then there is a regEx exception
+            try:
+                non_grad_updates = set(tf.get_collection(tf.GraphKeys.UPDATE_OPS), scope=scope)
```
ppwwyyxx: the parenthesis look wrong

PatWie:
Yeah. But the exception still happens iff the scope has no name.

ppwwyyxx:
I'm still not following why do you need the scope.
1. In tensorpack, `get_cost` was never called under a variable scope other than the default scope (with name == ''). In multi-tower training it is called under different name scopes, though.
2. In multi-tower training, some of my tests show that the regularization losses won't get repeatedly added to the collection, so there is not a need for de-duplication. You already uses `ctx.is_main_training_tower`, so UPDATE_OPS is well-handled in multi-tower setting.

BTW, looking at the TF code it looks like `tf.get_collection` accepts scope of string type.

ppwwyyxx:
Anyway, it should work already. I guess I can merge it first and look at the issues later.

ppwwyyxx:
At some time I stop summarizing "cost" by default, and this creates a problem with slim.
Because gradients doesn't actually depend on cost, so the UPDATE_OPS were never called.
I'll fix it soon.

ppwwyyxx:
I still thinks that users should handle the `REGULARIZATION_LOSSES` collection by themselves. Some reasons:
1. consistent with how regularization is done in tensorpack
2. It's not always easy to handle it automatically, especially when now we have many different multiGPU/distributed strategies I found it hard to make every scenario correct. Similar things happen to UPDATE_OPS as well. I believe anyone who writes pure tensorflow code would also have to take special care of these collections by themselves -- even I myself never trusted my own code and always tend to double check that I'm handling these collections correctly. 
3. Also, imagine using `tf.layers` inside GAN -- `get_cost` is never used, so you have to process the collection anyway. This looks like a surprise to users so perhaps it's better to be explicit than to do too much under the hood.

PR #82, opened 2017-01-02
Add example for extracting saliency using ResNet-50 with tfSlim

PatWie:
This is another example to merge from tfblueprint, which loads the Resnet-50 provided by Google for TensorFlow and computes the saliency map using guidedReLU. It also contains several other ways of visualizing outputs.


PatWie:
This requires #81 to be merge beforehand. I will rebase this pull-request on the master branch when #81 was merged.

PatWie:
Rebased on HEAD.

tensorpack/utils/utils.py
```diff
@@ -91,3 +93,62 @@ def get_tqdm(**kwargs):
...
+
+    Nice colormaps are:
+      - Blues, Blues_r
+      - Reds, Reds_r
+      - BuGns, BuGns_r
+      - Greys, Greys_r
+
+    Args:
+        intensity (TYPE): array of intensities such as saliency
+        background (None, optional): background for heatmap
```
ppwwyyxx: There isn't a background argument.

tensorpack/utils/utils.py
```diff
@@ -91,3 +93,62 @@ def get_tqdm(**kwargs):
...
+        cmap (str, optional): used colormap (required matplotlib)
+
+    Returns:
+        TYPE: nice heatmap
```
ppwwyyxx: Could you document the expected format of input/output image? Usually I will be especially confused with the range (0-1 or 0-255?). RGB/BGR is not a big deal in this case, though.

tensorpack/utils/utils.py
```diff
@@ -91,3 +93,62 @@ def get_tqdm(**kwargs):
...
+    import matplotlib.pyplot as plt
+
+    cmap = plt.get_cmap(cmap)
+    intensity = cmap(intensity.flatten())[..., 0:3].reshape([intensity.shape[0], intensity.shape[1], 3])
+    intensity = intensity[:, :, [2, 1, 0]]
+    return intensity
+
+
+def filter_intensity(intensity, rgb):
+    """Only highlight parts having high intensity values
```
ppwwyyxx: Could you move this function out to the example? I think different people always have different ideas about how to best blend two images for visualization, for their own tasks. For instance. they may want those regions with 0 intensity also partially visible. In one of my recent code [here](https://github.com/ppwwyyxx/Stereo-Pose-Machines/blob/master/src/model.py#L26) I used a different way of blending.

ppwwyyxx:
Another question I'm having in mind: could it be confusing to call it saliency ? This word never appears in the paper, and the author described the method as a visualization method to find out what the layers has learned. But "saliency" has its own field of research in CV.

PatWie:
Should I further make some edits? How? Amend previous commits, make new commits? I originally planned to do this tomorrow.
The correct term is: *saliency maps*. The best references are:

[Visualizing and Understanding Convolutional Networks](https://arxiv.org/pdf/1311.2901v3.pdf)
[Deep Inside Convolutional Networks: Visualising Image Classification Models and Saliency Maps
](https://arxiv.org/pdf/1312.6034v2.pdf)

The [guided backprob](https://arxiv.org/pdf/1412.6806v3.pdf) is based on these. Just being curious: Is there a reason to move the saliency_op from the symb-functions to this example.

I did a quick search on google and found a [related Theano version](https://github.com/Lasagne/Recipes/blob/master/examples/Saliency%20Maps%20and%20Guided%20Backpropagation.ipynb).

Btw: a related work is https://arxiv.org/pdf/1604.00825v1.pdf

ppwwyyxx:
I think I'll merge it after clean-up the README a bit. I planned to use fewer images and stack images together to leave space for other type of visualizations. For example I really like Fig.3 in https://arxiv.org/pdf/1412.6806v3.pdf.

I'm OK with the term saliency map, since the paper is using it. 
I moved the saliency function out because I was confused with the name myself. People could understand the function very easily by looking at the code alone but might get confused by looking at the name and document. Since it is well-defined in the literature we can move it back and explain it better.

tensorpack/utils/viz.py
```diff
@@ -242,6 +243,34 @@ def dump_dataflow_images(df, index=0, batched=True,
                 vizlist = vizlist[vizsize:]
 
 
+def intensity_to_rgb(intensity, cmap='jet', normalize=False):
```
PatWie: There was a reason for [not using JET](https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/) in my code.

ppwwyyxx: Wow thanks for pointing it out! I'm now using cubehelix as default.
Blues might be a boring default 😄

PR #90, opened 2017-01-06
More verbose information about dataflow

PatWie:
To verify the shapes of incoming data from the dataflow it is good
to actually output them for debugging reasons.

### DIFF
--- a/tensorpack/dataflow/common.py
+++ b/tensorpack/dataflow/common.py
@@ -4,6 +4,7 @@
 
 from __future__ import division
 import numpy as np
+from termcolor import colored
 from collections import deque, defaultdict
 from six.moves import range, map
 from .base import DataFlow, ProxyDataFlow, RNGDataFlow
@@ -12,7 +13,7 @@ from ..utils import logger, get_tqdm
 __all__ = ['TestDataSpeed', 'BatchData', 'BatchDataByShape', 'FixedSizeData', 'MapData',
            'MapDataComponent', 'RepeatedData', 'RandomChooseData',
            'RandomMixData', 'JoinData', 'ConcatData', 'SelectComponent',
-           'LocallyShuffleData']
+           'LocallyShuffleData', 'PrintData']
 
 
 class TestDataSpeed(ProxyDataFlow):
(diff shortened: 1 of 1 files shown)


PatWie:
That's way there is a try-catch-block. For most image-processing models it is good to double-check the dataflow-output on at least one sample.

ppwwyyxx:
Yes it would be good, but it's not general enough to make it a default inside trainer because there are cases it doesn't make sense.
You can implement it by adding a dataflow adapter, which prints the shape of the first datapoint. This way users can choose to use it, by e.g. (ds = PrintShape(ds)). This also avoid modifying code for every InputData type.

PatWie:
That's a cool idea.

PatWie:
The linting failed because the current master has one issue:

> ./ConvolutionalPoseMachines/load-cpm.py:27:1: E302 expected 2 blank lines, found 1

This was introduced in 13dc646b408e88532744c626691727a68e025bd0, see
https://github.com/ppwwyyxx/tensorpack/commits/master

ppwwyyxx:
Thanks for pointing out. It's working in HEAD now.

I still have a bunch of old code under my private examples/, so I haven't setup a pre-commit hook for it. Will make that later.

tensorpack/dataflow/common.py
```diff
@@ -471,3 +471,49 @@ def add_next():
...
+    def __init__(self, ds):
+        super(DebugData, self).__init__(ds)
+        self.ds = ds
+
+        ds.reset_state()
+        dummy = next(ds.get_data())
+        if isinstance(dummy, list):
+            logger.info("Incoming dataflow consists of %i elements" % len(dummy))
+            for k, entry in enumerate(dummy):
+                self.reason_input_data(entry, k)
```
ppwwyyxx: Could you move the implementation to `reset_state()`? `reset_state` of an underlying DataFlow only get called when you call reset_state from a high level DataFlow, this is currently true for all DataFlow except Prefetch.

I'm not sure will this cause any issues, though. I'm trying to think if I'm using this assumption somewhere.

tensorpack/dataflow/common.py
```diff
@@ -471,3 +471,49 @@ def add_next():
...
+            if hasattr(t, 'max'):
+                t_max = t.max()
+            if hasattr(t, 'min'):
+                t_min = t.min()
+
+            t_range = ("None-None")
+            if t_max is not None or t_min is not None:
+                t_range = "%.4f-%.4f" % (t_min, t_max)
+
+            print("%s dp %i: is %s of shape %s with range [%s]" % (" " * (depth * 2), k, t_type, t_shape, t_range))
```
ppwwyyxx: Could you use logger.info for the logging?

tensorpack/dataflow/common.py
```diff
@@ -471,3 +471,49 @@ def add_next():
...
+        if isinstance(dummy, list):
+            logger.info("Incoming dataflow consists of %i elements" % len(dummy))
+            for k, entry in enumerate(dummy):
+                self.reason_input_data(entry, k)
+
+    def reset_state(self):
+        self.ds.reset_state()
+
+    def size(self):
+        return self.ds.size()
```
ppwwyyxx: This is already implemented the same way in ProxyDataFlow.

tensorpack/dataflow/common.py
```diff
@@ -471,3 +471,49 @@ def add_next():
                     for v in self.q:
                         yield v
                     return
+
+
+class DebugData(ProxyDataFlow):
```
ppwwyyxx: Could you name it something more informative about what is happening? Such as PrintData/LogData/PrintDataShape/PrintDataInfo, similar to TestDataSpeed I'm having.
PrintData might not be very bad because in the future, options can be added to print k datapoints instead of just 1.

tensorpack/dataflow/common.py
PatWie: done like in ModelDump.

tensorpack/dataflow/common.py
PatWie: If I would put this into `reset_state()` then there would be no output in
````python
def get_data():
    ds = CaffeLMDB('path/to/lmdb')
    ds = SomeInscrutableMappings(ds)
    ds = DebugData(ds, num=2)
    return ds
ds = get_data()
````

ppwwyyxx:
Thanks a lot and especially for the detailed documentation!

tensorpack/dataflow/common.py
ppwwyyxx: The protocol is that reset_state should always get called before a DataFlow start producing. But of course putting it in __init__ can make it print the log earlier and probably that's more preferable for the debugging purpose.

Issue #39, opened 2016-10-31
Before and After Step Design


pawelsubko:
Hi,
Could you please alow acces to "before_step" and "after_step" ? sometimes it is crucial to 
1. run something more often then once per epoch (or just with different frequency) 
2. to run something explicitely BEFORE train_op (like some tensor statistics)
3. it makes life easier defining timers etc. (it is crucial to benchmark times of train_op, other_run_ops, loading data etc.)

ppwwyyxx:
Sure it won't be hard to make the changes. But meanwhile I'm also curious about what your use case is and see if there's a better solution for that.
1. To run things more frequently you can change the epoch size, or use `PeriodicCallback` to adjust the frequency. Unless you really want something to run very frequently (every several iterations), and I wonder what that could be.

2. What do you mean by tensor statistics? `before the train_op` seems no different from `after the last train_op`, unless you're using the less efficient SimpleTrainer which loads data between two calls to the op.

3. As said above, you cannot accurately test the time of data loading by this two callbacks, because it is supposed to run in parallel to train_op. 

To test the speed of your dataflow, you can either write a small function which simply loops over the dataflow, or do `TestDataSpeed(dataflow, size=1000).start_test()` which does the same thing.

The iteration speed given in training should be the speed of `train_op`, unless the data is slower, in which case you can use `FakeData` instead to obtain a speed measurement of `train_op`. And what do you mean by `other_run_ops`?

pawelsubko:
0. That frequently? Just um... (it's complicated) sending some data to the server. Sending it every step is not the best solution.
I. Ok, I may completely miss the point. Could you help me on the following problem:
What is a best way to compute loss on a given datapoint, then run train_op on this datapoint? (otherwise it's a waste!)
II. The way I "normally" do is I benchmark the whole loop body including the data_generator.next() call and all the other calls, and the train step only to get reliable (?) benchmark whether GPU train_op or some other calls/cpu<->gpu transfers etc. are the bottleneck. (Last time inefficiently written Augmentation (?) caused 1s whole loop vs 450ms train_op, the problem may be somewhere else tho)

ppwwyyxx:
I. I'm still not sure.. do you mean "if the loss is small enough, then don't train on this data point"?
If that is the case, you can probably just implement this in the graph, with tf.cond and tf.stop_gradient. But  maybe doing something on dataflow to not feed such datapoint would be nicer.

[...]

PR #91, opened 2017-01-06
Add Livedata

PatWie:
livemeter
Sometimes it is necessary to inspect scalar values within an epoch.
This commit replaces tqdm with a modification to support livedata.

This should also be benchmarked a large-scale models. In simple MNIST examples I get no significant overhead.

### DIFF
--- a/examples/SimilarityLearning/mnist-embeddings.py
+++ b/examples/SimilarityLearning/mnist-embeddings.py
@@ -130,13 +130,17 @@ class SoftTripletModel(TripletModel):
         return symbf.soft_triplet_loss(a, p, n, scope="loss")
 
 
-def get_config(model):
+def get_config(model, algorithm_name):
     logger.auto_set_dir()
 
     dataset = model.get_data()
     steps_per_epoch = dataset.size()
 
     lr = symbf.get_scalar_var('learning_rate', 1e-4, summary=True)
+
+    extra_display = ["cost"]
+    if not algorithm_name == "cosine":
+        extra_display = extra_display + ["loss/pos-dist", "loss/neg-dist"]
 
     return TrainConfig(
         dataflow=dataset,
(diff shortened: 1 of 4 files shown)


ppwwyyxx:
There is a plan to more generally support `trigger_step()` with user-specified data. Some discussion were in #39. The idea is that a Callback should define what it needs every step and what to do with it, not only to print them.

PatWie:
But the problem with any before/after step design is that this needs explicit addition `sess.run`. Here in this implementation it is simply extracted from the graph, when it is already computed.

ppwwyyxx:
It will not need an extra sess.run. The trainer should know what all callbacks need, and fetch those tensors together with `train_op` in one sess.run.

PatWie:
So you mean, for some basic statistics of scalars you prefer callbacks instead of simple tf.collections? Is was first thinking about add this functions to `TrainerConfig` with regEx.
It seems like flake8 is really bothersome wrt. to xrange, though it is caught in the correct way.

ppwwyyxx:
In fact I'd prefer no logic been implemented inside trainer if they can be done outside. The current epoch-wise logging is also done outside.

PatWie:
I understand the current design. But within epochs you need to pass the data to tqdm or midifications of it. Another idea would be a TrainingStepCallback which contains the code of the most-inner loop from the training-main-loop. This would probably also help to change training-logic to GAN training. What do you think?

ppwwyyxx:
I can consider an option to disable tqdm, so that users can have more control over what to display within steps.

ppwwyyxx:
Please leave the PR as-is or close it for now. We'll revisit it and can adopt the live data monitor when trigger_step is ready.

PatWie:
Any clever idea to adapt this feature as a step-trigger? Overriding just the default? Or replacing the tqdm callback?

ppwwyyxx:
Now that tqdm is a callback, this can probably be implemented as a callback as well. And then users get to choose what to use.

ppwwyyxx:
Do you think [this tqdm feature](https://github.com/tqdm/tqdm#description-and-additional-stats) works for the case here?

PatWie:
I am not sure if the progress bar really has to be a progress.

ppwwyyxx:
What do you mean?

PatWie:
@ppwwyyxx very nice refactoring! 
I added these extra_fetches to Progressbar directly using postfix.

It would be really cool, if tqdm could display both: examples/sec and sec/example.

https://github.com/tqdm/tqdm/issues/341

Issue #92, opened 2017-01-06
symbolic_functions is not defined


PatWie:
````
~/tmp/tensorpack/examples (master) $ python mnist-convnet.py 

Traceback (most recent call last):
  File "mnist-convnet.py", line 168, in <module>
    SimpleTrainer(config).train()
  File "/home/wieschol/git/tensorpack/tensorpack/train/base.py", line 58, in train
    self.setup()
  File "/home/wieschol/git/tensorpack/tensorpack/train/base.py", line 126, in setup
    self._setup()
  File "/home/wieschol/git/tensorpack/tensorpack/train/trainer.py", line 83, in _setup
    model.build_graph(self.input_vars)
  File "/home/wieschol/git/tensorpack/tensorpack/models/model_desc.py", line 108, in build_graph
    self._build_graph(model_inputs)
  File "mnist-convnet.py", line 87, in _build_graph
    wrong = symbolic_functions.prediction_incorrect(logits, label, name='incorrect')
NameError: global name 'symbolic_functions' is not defined
````

possible bugfix
[bugfix.txt](https://github.com/ppwwyyxx/tensorpack/files/690768/bugfix.txt)

edit: same for `summary`

PatWie:
Possible patch

````git
From e8f77a7918eada5505a92d89d4b8476f0257eb23 Mon Sep 17 00:00:00 2001
From: PatWie <patrick@wieschollek.info>
Date: Fri, 6 Jan 2017 21:27:04 +0100
Subject: [PATCH] fix

---
 tensorpack/tfutils/__init__.py | 8 +++++---
 1 file changed, 5 insertions(+), 3 deletions(-)

diff --git a/tensorpack/tfutils/__init__.py b/tensorpack/tfutils/__init__.py
index f75d947..57fbb11 100644
--- a/tensorpack/tfutils/__init__.py
+++ b/tensorpack/tfutils/__init__.py
@@ -9,6 +9,7 @@ __all__ = []
 
 
 def _global_import(name):
+    __all__.append(name)
     p = __import__(name, globals(), None, level=1)
     lst = p.__all__ if '__all__' in dir(p) else dir(p)
     for k in lst:
@@ -21,12 +22,13 @@ _TO_IMPORT = set([
     'common',
     'gradproc',
     'argscope',
-    'tower'
+    'tower',
+    'symbolic_functions',
+    'summary'
     ])
 
 _CURR_DIR = os.path.dirname(__file__)
[...]

PR #93, opened 2017-01-06
Travis should run some examples

PatWie:
Sorry for another pull-request. To make sure issues like #92 do not happen, this adds a simple unit test to run a script for some seconds.

This is just a pull-request to actually test the pull-request itself.

To use this feature it is a good idea to enable the [branch protection from GitHub](https://github.com/blog/2051-protected-branches-and-required-status-checks) for everybody :wink: 

TODO:

- [ ] install opencv2
- [ ] cache dataset folder
- [ ] other useful examples to run?
- [ ] clean-up `trainlog_dir` when testing multiple examples
- [ ] deploy documentation?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,46 @@
-language: "python"
-python:
-   - "2.7"
-   - "3.5"
 sudo: false
-cache: pip
+dist: trusty
+language: "python"
+
+cache:
+  pip: true
+  apt: true
+
+addons:
+    apt:
+        sources:
+            - ubuntu-toolchain-r-test
+        packages:
+            - g++-4.8
+            - python-dev
+            - python-opencv
+            - libopencv-dev
+
+matrix:
+    fast_finish: true
+    include:
+        - os: linux
+          python: 2.7
+          env: TF_BINARY_URL=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.12.1-cp27-none-linux_x86_64.whl
+        - os: linux
+          python: 3.5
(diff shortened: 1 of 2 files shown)


PatWie:
It seems to be nearly impossible to run TensorFlow(CPU) and OpenCV with Travis-CI. I will reopen this pull-request, when I found a solution.

ppwwyyxx:
Why does the import error happen?

PatWie:
That's the point. There are several issues with Python under Travis:
- NumPy uses ucs2 and TensorFlow ucs4
- there is no OpenCV package for Python3
- there seem to be inconsistent Python versions of 3.4 within a single Travis-CI container
- compiling OpenCV from source takes quite some time (6min)

I will further try to debug the process here:
https://github.com/PatWie/travis_python_cv2_tf

which is a cleaned version of my 75 commits from yesterday:
https://github.com/PatWie/python_cv2_test

Unfortunately, running the Travis-containers locally gives not the same environment and same paths. Even worse, the "Trusty" containers are in beta. So if they change their stuff as frequently as TF does, it will be tedious to adapt these changes.

Maybe some of these might be worth to look at:
- https://circleci.com/features/
- http://codeship.com/

I would be happy if even a combination of Python2.7+TensorFlow0.12+OpenCV3 would work.

PatWie:
I switched to CircleCI which saved a lot of time. 

#### Outputs:
with issue #92 
https://circleci.com/gh/PatWie/tensorpack-fork/5#config/containers/0

#### with current HEAD+this pull-request
https://circleci.com/gh/PatWie/tensorpack-fork/7

This nice thing is that the entire run only takes bout 2minutes.

This also reveals that there might be another problem, although it is running fine here.

PR #98, opened 2017-01-11
Add continous integration

PatWie:
Use Travis-CI to run some tests under the latest release of TensorFlow and nightly releases in python 2.7 and 3.5.

This is a clean version of #93, which cannot be opened again because of rebasing to master.

features:
- TensorFlow version is environment variable
- run tests with the latest nightly build of the current version to check changes in API (these are allowed to fail)
- run arbitrary Python scripts with a timeout to test imports

for output see:
https://travis-ci.org/ppwwyyxx/tensorpack/builds/191097681

This further allows to:
- add tests for dataflow from this lib
- add python [unit-tests](https://docs.python.org/2/library/unittest.html)
- automatically [deploy Sphinx-doc](https://drdoctr.github.io/doctr/)
- automatically [deploy pip package](http://5minutes.youkidea.com/howto-deploy-python-package-on-pypi-with-github-and-travis.html)

It might also be possible to cache the MNIST dataset actually to run the example (including the computation graph). I doubt that this is no possible at the moment within the current timeout settings as it has to download the dataset.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,49 @@
-language: "python"
-python:
-   - "2.7"
-   - "3.5"
 sudo: false
-cache: pip
+dist: "trusty"
+language: "python"
+
+cache:
+  pip: true
+  apt: true
+
+addons:
+    apt:
+        sources:
+            ubuntu-toolchain-r-test
+matrix:
+    fast_finish: true
+    include:
+        - os: linux
+          python: 2.7
+          env: TF_VERSION=0.12.1 TF_TYPE=release
+        - os: linux
+          python: 3.5
+          env: TF_VERSION=0.12.1 TF_TYPE=release
+        - os: linux
+          python: 2.7
+          env: TF_VERSION=0.12.1 TF_TYPE=nightly
+        - os: linux
+          python: 3.5
+          env: TF_VERSION=0.12.1 TF_TYPE=nightly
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,19 +1,46 @@
...
+    - if [ -e requirements.txt ]; then pip install -r requirements.txt;else pip install -r requirements.pip;fi
+    - ./tests/install-tensorflow.sh
 
 before_script:
-  - flake8 --version
+    - flake8 --version
 
 script:
-  - flake8 .
-  - cd examples && flake8 .
```
ppwwyyxx: Why did you remove the linting for examples?

PatWie: It should also be linted maybe you should fix that afterwards. But flake8 does its job recursively.

ppwwyyxx: Yes but I ignored examples in /tox.ini, and examples/tox.ini is different from tox.ini (ignore certain type of errors).

PatWie: I understand. I will fix it and also print the version of tensorflow and opencv.

PatWie:
I change those thing and added
````bash
python -c "import cv2; print('OpenCV '+ cv2.__version__)"
python -c "import tensorflow as tf; print('TensorFlow '+ tf.__version__)"
````

PR #102, opened 2017-01-15
Add example for several embedding methods

PatWie:
This also includes an animate script using avconv and app-util to
reduce repeating code (DRY).

Are you interested in app.py ?


PatWie:
@ppwwyyxx 
Do you know what is going on with the python3 version [here](https://travis-ci.org/ppwwyyxx/tensorpack/jobs/192083670). I only use python2.

ppwwyyxx:
Don't worry that's a bug of mine, should be fixed already.

ppwwyyxx:
Thanks! I'll run it some time to figure out what's happening. Some general comments:
1. I use --gpu everywhere only because it's shorter than CUDA_VISIBLE_DEVICES, but I hope not to provide a set of predefined "standard" arguments, at least not now. That reminds me of caffe where I'll need to make some changes in code, some in config file, some in command line.
It will look better to me if the interface is different, such as one line of `enable_common_flags()`, to make it clearer this is optional. But one single line of `os.environ['CUDA_VISIBLE_DEVICES']` is probably not worth the complexity.
2. There is no need to use variable_scope and the reuse option, because there are no variables defined in those functions. (If there are, I would put them in `models/` instead of `symbolic_functions`.) You probably want `tf.name_scope`.
3. the argscope from slim would make things complicated because I had an argscope for models/ already. And seem like you're not using them anyway.
I think argscope only helps when you need to call a function with __long argument list__ for __multiple times__. This is usually not true for what's in `symbolic_functions`.
4. animate.py introduces too many dependencies, but putText and VideoWriter in cv2 should be enough (may not support gif format specifically, though). But if I were to write it, I'll just write one function which produces animation from a iterator of images. I wouldn't assume people want to put text at (0,0).
Also, if you just use it to create animation from files, ffmpeg might be simpler.

PatWie:
I will apply these changes 1-3.
Unfortunately, there is no way using OpenCV for animated gifs. But I can remove the animated-gif script. I though the `scripts` folder is more or less a collection of scripts.

PatWie:
1. `change_gpu` is really nice!
2. fixed. I added variables during playing around with these loss-functions
3. it's a bad habit of mine
4. I removed it. But it would be a nice and helpful script. Avconv is the replacement for ffmpeg.

examples/SimilarityLearning/mnist-embeddings.py
```diff
@@ -0,0 +1,319 @@
...
+    embed = np.zeros((BATCH_SIZE * NUM_BATCHES, 2))  # the actual embeddings in 2-d
+
+    # get only the embedding model data (MNIST test)
+    ds = EmbeddingModel.get_data()
+    ds.reset_state()
+
+    offset = 0
+    for digit, label in ds.get_data():
+        # TODO: investigate why [0][0]
+        prediction = pred([digit])[0][0]
```
PatWie: I do not really get why `[0][0]` works instead of `[0]`. Isn't it `[component-id][batch-entry]` or did I mess up the dataflow?

ppwwyyxx: pred() return a list of outputs. You only want 1 output tensor so it's a list of length 1.
It's `[output tensor id][batch entry]`.

tensorpack/tfutils/symbolic_functions.py
```diff
@@ -212,3 +212,142 @@ def saliency_map(output, input, name="saliency_map"):
     saliency_op = tf.gradients(max_outp, input)[:][0]
     saliency_op = tf.identity(saliency_op, name=name)
     return saliency_op
+
+
+def l2_norm(t, eps=1e-12):
+    """return L2 norm for input x
```
ppwwyyxx: I don't think reducing on axis=1 (not axis=None, not axis=[1,2,...end], but axis=1) is what people would expect on a function named `l2_norm`. And it would also take more documentation to explain that it reduces on axis=1, than writing the code itself.
Could you just put this function where it is used?

tensorpack/tfutils/symbolic_functions.py
```diff
@@ -212,3 +212,142 @@ def saliency_map(output, input, name="saliency_map"):
...
+        pos_dist = tf.reduce_mean(tf.sqrt(d_pos + 1e-6))
+        neg_dist = tf.reduce_mean(tf.sqrt(d_neg + 1e-6))
+
+        if extra:
+            return loss, pos_dist, neg_dist
+        else:
+            return loss
+
+
+def soft_triplet_loss(a, p, n, eps=1e-6, extra=True):
```
ppwwyyxx: It's not consistent when some functions in this file accept eps as an option but some just use a constant.

Also, could you use a more descriptive name rather than a,p,n?

examples/SimilarityLearning/mnist-embeddings.py
```diff
@@ -0,0 +1,315 @@
...
+        x, y = self.embed([x, y])
+        # tag them, just for inference later on
+        x = tf.identity(x, name="emb")
+
+        # compute the actual loss
+        cost, pos_dist, neg_dist = symbf.contrastive_loss(x, y, label, 5., extra=True)
+        self.cost = tf.identity(cost, name="cost")
+
+        # track these values during training
+        tf.summary.scalar('pos_dist', pos_dist)
```
ppwwyyxx: summary op only gets run once after each epoch (and on the last data point only). For scalar summaries you probably want to use `add_moving_summary`, which computes moving average over every iteration.

tensorpack/callbacks/saver.py
```diff
@@ -78,6 +78,7 @@ def _trigger_epoch(self):
                 self.path,
                 global_step=get_global_step(),
                 write_meta_graph=False)
+            logger.info("Model saved to %s" % (self.path))
```
ppwwyyxx: Maybe you'd like self.checkpoint_dir better?
self.path would look like "train_log/dirname/model". It's just a prefix of where the actual model is saved.

Or you can use `tf.train.get_checkpoint_state(self.checkpoint_dir).model_checkpoint_path` to get the path to the actual model.

ppwwyyxx:
About the naming in examples, I started to use ucfirst for directories simply because I found a lot of them starts with abbreviations.. like HED, GAN, A3C, DQN. So I used uppercase for all of them.
Other files are lower-case unless they need to be imported (then it has to be lower_case).

ppwwyyxx:
About `pos_dist` and `neg_dist`. Is it a good idea to use ``reduce_mean(y * dist)``? Wouldn't it be better to use ``reduce_sum(dist) / tf.count_nonzero(y)``?
Might not make a big difference when 0/1 labels are perfectly balanced, though.

ppwwyyxx:
Another question, is there a reason why eps is different? I saw 1e-12, 1e-10 and 1e-6 in the file.

tensorpack/tfutils/symbolic_functions.py
```diff
@@ -212,3 +212,149 @@ def saliency_map(output, input, name="saliency_map"):
...
+
+        delta = tf.reduce_sum(tf.square(diff), 1)
+        delta_sqrt = tf.sqrt(delta + 1e-6)
+
+        match_loss = delta
+        missmatch_loss = tf.square(tf.nn.relu(margin - delta_sqrt))
+
+        loss = tf.reduce_mean(0.5 * (y * match_loss + (1 - y) * missmatch_loss))
+        pos_dist = tf.reduce_mean(y * delta_sqrt, name="pos-dist")
+        neg_dist = tf.reduce_mean((1 - y) * delta_sqrt, name="neg-dist")
```
ppwwyyxx: Is it a good idea to use ``reduce_mean(y * dist)``? Wouldn't it be better to use something like ``reduce_sum(y * dist) / tf.count_nonzero(y)``?
Might not make a big difference when 0/1 labels are perfectly balanced, though.

tensorpack/tfutils/symbolic_functions.py
```diff
@@ -212,3 +212,149 @@ def saliency_map(output, input, name="saliency_map"):
...
+            tf.Tensor: norm of input tensor
+        """
+        with tf.name_scope("l2_norm"):
+            return tf.sqrt(tf.reduce_sum(tf.square(t), 1, True) + eps)
+
+    with tf.name_scope("cosine_loss"):
+        y = 2 * tf.cast(y, tf.float32) - 1
+        pred = tf.reduce_sum(left * right, 1) / (tf.squeeze(l2_norm(left) * l2_norm(right)) + 1e-10)
+
+        return tf.nn.l2_loss(y - pred)
```
ppwwyyxx: So cosine_loss is summed over the batch, but others are averaged over the batch? This is inconsistent and will confuse people when they want to interpret the magnitude of loss.

tensorpack/tfutils/symbolic_functions.py
```diff
@@ -212,3 +212,149 @@ def saliency_map(output, input, name="saliency_map"):
...
+        if extra:
+            return loss, pos_dist, neg_dist
+        else:
+            return loss
+
+
+def soft_triplet_loss(anchor, positive, negative, extra=True):
+    """Loss for triplet networks
+
+    same as Triplet Network Loss but without explicit margin
```
ppwwyyxx: Could there be either a formula or a reference to a paper explaining what it actually computes?

tensorpack/tfutils/symbolic_functions.py
ppwwyyxx: Yeah I noticed that I had a problem like this in another function and just fixed it in eb1fa920e81e32d8da5a45ede9b.
For this case, maybe return 0? Or maybe assertion?

PatWie: I would stick with the mean as it works even on large scale problems and this is how it is implemented [in Caffe](https://github.com/BVLC/caffe/blob/master/src/caffe/layers/contrastive_loss_layer.cpp#L60).

ppwwyyxx: I'm talking about pos_dist and neg_dist (only used for summary), not the loss.

PatWie: You mean
````python
pos_dist = tf.where(tf.equal(tf.reduce_sum(y), 0.0), 0, tf.reduce_sum(y * delta_sqrt) / tf.count_nonzero(y))
````

**Comment by PatWie** [comment]
Did you turned Travis-CI off?

File: tensorpack/tfutils/symbolic_functions.py [hunk 9]
```diff
@@ -301,22 +292,23 @@ def triplet_loss(anchor, positive, negative, margin, extra=False):
...
-    Learn embeddings from an anchor point and a similar input (p) as well as a not similar input (n)
-    Intuitively, a matching-pair (anchor, positive) should have a smaller relative distance
+    Learn embeddings from an anchor point and a similar input (positive) as
+    well as a not-similar input (negative).
+    Intuitively, a matching pair (anchor, positive) should have a smaller relative distance
     than a non-matching pair (anchor, negative).
 
     .. math::
-        \min \max(0, m + \Vert a-p\Vert^2 - \Vert a-n\Vert^2)
+        \max(0, m + \Vert a-p\Vert^2 - \Vert a-n\Vert^2)
```
PatWie: it is "min". Note "\Vert a-p\Vert^2" should be small and  " **-** \Vert a-n\Vert^2 " should be small

ppwwyyxx: The function only documents what it computes -- max(0, m+...)
\min is how it should be used (minimized), and this is literally what "loss" means.

PatWie: Ahh. Maybe we should write "tf.nn.relu(.....)".

ppwwyyxx: You can. They look the same to me. maybe will have some very slightly performance difference?

PatWie: I think the current version is ok

Issue #105, opened 2017-01-15
Improve GAN Network 


Skylion007:
Hello, let me just say I am a big fan of the code in this repo. I noticed though that the reference implementation you use for the GAN example is from November of 2015. I'd like to suggest that the code be improved by utilizing the latest techniques from a recent OpenAI [paper](https://arxiv.org/pdf/1606.03498v1.pdf). The code can be found [here](https://github.com/openai/improved-gan) and is already in Tensorflow, so it shouldn't be too hard to import. It would be a welcome addition since the main reason I use this project is for the very efficient GAN implementation. From what I've read from skimming over the paper, OpenAI modifies the loss function for the generator and discriminator to make them converge faster.

PatWie:
I have implemented these changes in my version:
- mini-batch discrimination
- switched the labels 0 <-> 1
- label-smoothing

However, I did not observe any (dramatic) changes in the examples, did you run some tests? 

But they have the correct *log*-likehood term in InfoGAN :wink:  I don't know what @ppwwyyxx's plans are. But the next example I planned to move to tensorpack is InfoGAN with a distribution-class for different codes.

Skylion007:
@PatWie I've noticed some difference on my own dataset with improved GAN. It seemed to help with my dataset in terms of yielding qualitatively better results at earlier epochs on their implementation. I maybe mistaken but isn't one important part of the paper you are missing is feature matching in section 3.1? Or have you called it something else? It also sounds like it will fix the issue you are having currently where the GAN is unstable. That sounds like possibly the most important improvement in the paper, but I could be mistaken.

ppwwyyxx:
About improved GAN, it interests me as well because it introduces a lot of new ways of training a model, so it would be good to implement them to test whether the design of tensorpack is really flexible.

About InfoGAN, @PatWie could you elaborate what do you mean by "correct log-likelihood"? I used OpenAI code as a reference when I wrote the examples but maybe I missed something.
Also it would be nice to have distribution classes like the original code.

PatWie:
> correct log-likelihood
I didn't saw the [sign](https://github.com/ppwwyyxx/tensorpack/blob/master/examples/GAN/InfoGAN-mnist.py#L83).

Skylion007:
[...]

PR #107, opened 2017-01-16
GAN refactor

PatWie:
This is an attempt on #105 but will take some more commits as I have to sort out my code. It removes unnecessary duplicated code and enhances TensorBoard outputs (plots and graph).

- [x] grouping tf.ops
- [x] separate latent factors code from model to support more distributions
- [x] adapt `sample`


examples/GAN/GAN.py
```diff
@@ -7,11 +7,67 @@
...
+        Returns
+        -------
+        TYPE
+            Description
+        """
+        all_vars = tf.trainable_variables()
+        self.g_vars = [v for v in all_vars if v.name.startswith('gen/')]
+        self.d_vars = [v for v in all_vars if v.name.startswith('discrim/')]
+
+    def build_losses(self, logits_real, logits_fake, reuse=False):
```
ppwwyyxx: reuse is not used?

tensorpack/tfutils/distributions.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _loglikelihood(self, x, theta):
+        eps = 1e-8
+        return tf.reduce_sum(tf.log(theta + eps) * x, reduction_indices=1)
+
+    def _prior(self):
+        return tf.constant([1.0 / self.cardinality] * self.cardinality)
+
+    def _code(self, batch_size, name="zc"):
+        ids = tf.multinomial(tf.zeros([batch_size, self.cardinality]), num_samples=1)[:, 0]
```
ppwwyyxx: Should "code" be renamed to "sample"? That's essentially what it is doing.
"code" is a name used particularly in InfoGAN.
Also, the placeholder_with_default hack is used in InfoGAN only to make inference work. It should stay in InfoGAN but not as a default behavior in distributions.

tensorpack/tfutils/distributions.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @class_scope
+    def mutual_information(self, x, theta):
+        expected_log_qc = self.entropy(x)
+        expected_log_qc_given_x = self.cross_entropy(x, theta)
+        return tf.add(expected_log_qc, expected_log_qc_given_x, name="mutual_information")
+
+    @class_scope
+    def entropy(self, x):
+        return tf.reduce_mean(self.loglikelihood_prior(x), name="entropy")
```
ppwwyyxx: Regardless of what is actually used for training, this is a "distribution" class so it should follow strictly the mathematical definition. Entropy should have a negative sign.

tensorpack/tfutils/distributions.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @class_scope
+    def cross_entropy(self, x, theta):
+        return tf.reduce_mean(self.loglikelihood(x, theta), name="cross_entropy")
```
ppwwyyxx: This as far as I know is an __approximation__ of the cross entropy (that is used in InfoGAN specifically but not in general(?)). Is it really a good idea to put it here?

PatWie: Yes, the mutual information can only be approximated by a lower bound on the cross entropy.

tensorpack/tfutils/distributions.py
PatWie: Yeah, the placeholder should get sample as the default only. I will adjust that after the product-distribution class is finished.

PatWie:
@ppwwyyxx Can you have a look before at the InfoGAN before I adjust the other examples.

graph

I would say it definitely learns some latent uniform distribution

latent

loss



But it is not what I expected.

ppwwyyxx:
The uniform distribution doesn't look as good as the papers (from figure and curve), is that what you mean?

The current code looks good, thanks a lot for that! Could you put the changes related to improved GAN in a separate PR? Then this PR would be only about InfoGAN and easier to manage.
The other PR might depend on GANModelDesc so I'll try to merge this soon.

PatWie:
> The uniform distribution doesn't look as good as the papers (from figure and curve), is that what you mean?

Yeah, I do not get why my results are never as smooth as in the papers :cry: 

> The current code looks good, thanks a lot for that! Could you put the changes related to improved GAN in a separate PR?

The only change mentioned in #105 is currently "swap labels". But anyway when merging this one needs to update image2image, ... a well.

ppwwyyxx:
There are some problems I found:
* Naming confusions. Comparing to the original InfoGAN code, your "sample" is actually their "sample_prior", and they do have a "sample" method that does "sampling from a batch of dist_info" as one could expect from the name. Your "entropy" is actually "entropy under the prior distribution" but their "entropy" is different (and not used). "activate_dist" was OK (it performs an activation) but "model_param" is probably more confusing (?).
* You probably can check the static shape of input/output in the base class. For example, prior()  should return some thing with shape "param_dim". This would be very helpful especially when now it's confusing what these methods really mean.
* They have "prior_dist_info", and a "sample" method which takes a batch of "dist_info". Combining the two would give a "sample_prior" which is your "sample". This naturally solves the 1D/2D shape problem you're having as the TODO, as you can then require loglikelihood to take 2D input.
* From their code the "prior" of Uniform should be zero-mean, unit-stddev, and "sample_prior" is override to not sample from this prior. The prior in your code is different, and this will affect how "entropy under prior" is computed. Perhaps this explains your results.

PatWie:
Well, I haven't looked at their code in detail. I just wrote a LaTeX file with all ugly calculations. In the end the entire InfoGan is simply a weighted sum of mutual information (including the GAN objective). Maybe I missed something. I will check it later.

PatWie:
This gives

infogan-mnist

The issue with the "7" and "9" columns can be probably solved by other hyper-parameters or even another run.

ppwwyyxx:
Yes even the figure in the paper has some 7 vs 9 confusion.

PatWie:
You accidentally used an old Readme during merging ;-)

ppwwyyxx:
I just noticed this line:
```
        Note, we swap 0, 1 labels as suggested in "Improving GANs".
```
What does it mean? I only know one trick used here: min log (1-D) -> max(log(D)), and this is suggested by the original GAN paper.

PatWie:
I somewhere read it. But I never notice any differences. It is probably better to flip these labels real(1) and fake(0).

Skylion007:
@ppwwyyxx Here is the relevant paper: https://arxiv.org/pdf/1606.03498v1.pdf

PatWie:
They talk about label-smoothing, not flipping.

ppwwyyxx:
@PatWie  Is there a reference? It doesn't sound like it will work -- it sounds equivalent. You let the discriminator predicting P(sample is fake) instead of P(sample is real) = 1 - P(sample is fake), but sigmoid(x) = 1 - sigmoid(-x) and weights are symmetrically initialized, so predicting either of them is the same.

Skylion007:
> I have implemented these changes in my version:
> 
>     mini-batch discrimination
>     switched the labels 0 <-> 1
>     label-smoothing
> 
@PatWie 

Second comment on #105

PatWie:
Yeah, there is no difference. I think I misinterpreted
> replaces the 0 and 1 targets for a classifier with smoothed values, like .9 or .1

I would go with the default formulation. This affects almost everything `ones_like, zeros_like, <, >` in
https://github.com/ppwwyyxx/tensorpack/blob/master/examples/GAN/GAN.py#L50-L64

Sorry. It should be
````python
with tf.name_scope("GAN_loss"):
    score_real = tf.sigmoid(logits_real)
    score_fake = tf.sigmoid(logits_fake)
    tf.summary.histogram('score-real', score_real)
    tf.summary.histogram('score-fake', score_fake)

    with tf.name_scope("discrim"):
        d_loss_pos = tf.reduce_mean(tf.nn.sigmoid_cross_entropy_with_logits(
            logits=logits_real, labels=tf.ones_like(logits_real)), name='loss_real')
        d_loss_neg = tf.reduce_mean(tf.nn.sigmoid_cross_entropy_with_logits(
            logits=logits_fake, labels=tf.zeros_like(logits_fake)), name='loss_fake')

        d_pos_acc = tf.reduce_mean(tf.cast(score_real > 0.5, tf.float32), name='accuracy_real')
        d_neg_acc = tf.reduce_mean(tf.cast(score_fake < 0.5, tf.float32), name='accuracy_fake')

        self.d_accuracy = tf.add(.5 * d_pos_acc, .5 * d_neg_acc, name='accuracy')
        self.d_loss = tf.add(.5 * d_loss_pos, .5 * d_loss_neg, name='loss')

    with tf.name_scope("gen"):
        self.g_loss = tf.reduce_mean(tf.nn.sigmoid_cross_entropy_with_logits(
            logits=logits_fake, labels=tf.ones_like(logits_fake)), name='loss')
        self.g_accuracy = tf.reduce_mean(tf.cast(score_fake > 0.5, tf.float32), name='accuracy')

    add_moving_summary(self.g_loss, self.d_loss, self.d_accuracy, self.g_accuracy)
````

PR #123, opened 2017-01-28
super-resolution

PatWie:
I hope, it is ok to make an early pull-request although there are still some issues.
I would further propose to remove `psnr_loss` and simply add `psnr` (negative of psnr-loss).

@ppwwyyxx is there an issue with the prefetching? I get
````
InvalidArgumentError (see above for traceback): Input to reshape is a tensor with 43350 values, but the requested shape requires a multiple of 65025
````

and I really expect a multiple of `65025 = 85 * 3 * 85 * 3`


ppwwyyxx:
The rename sounds good. It's more straightforward that way.

The error is from tf.reshape so it seems to suggest something wrong with the symbolic code.

PatWie:
Ok it works. I will push again, when the model is trained and I have some results. Can you please change the `psnr` with your next commit?

ppwwyyxx:
Renamed in 6e24b953218dc.

PatWie:
The network model and parameters are the same as in the paper. However, there are some difference in the pre-processing:
- in their work they use YCbCr colorspace (should not make differences when using large batchsize)
- they first blur and then sub-sample for low-res, here I simply use bilinear downsampling
- for their final results they trained in ImageNet, this implementation just uses BSDS500 (I should download ImageNet sometime

**Warning:** I cannot replicate the quality which is mentioned in the paper. This probably requires longer training and a larger dataset.

ppwwyyxx:
why is this closed?

PatWie:
Ah, I was cleaning unused branches and accidentally delete this branch. Sorry for that.

PatWie:
I trained this network on ILSVRC12 for several days with Adam as in the paper and a scheduled LR (which gave better results). However, these are still far away from the results in the paper. I think this should be just a demo for training on small images and applying the arbitrary images.

The PixShift layer has been checked manually.

PatWie:
Cannot reproduce results even when playing with hyperparameters. Still it is nice example to show how to use fully-convolutional networks during inference.

ppwwyyxx:
I also think it would be a nice example if the results can be better (now it seems worse than bicubic?) And I like the SuperResolution topic so I may try it when I have more time.

PatWie:
Superresolution is more or less a hobby for me. So I do not really rely on this. And the result is really worse than bicubic (I tried to optimize MSE or PSNR directly).

Issue #129, opened 2017-02-01
Bug OpenCV does not preserve image shape of grayscale images.


Skylion007:
When I call imaug.GaussianBlur on grayscale images, instead of returning (batch_size, 256, 256, 1) it truncates the grayscaling and just returns an image of shape (batch_size, 256, 256). This is inconsistent with imaug.crop() and imgaug.resize(). I solved my issue by manually resizing the image to be the same dimensions of the input image. I suspect this is an issue with OpenCV in general though, so imaugs that use OpenCV should make sure to explicitly restore the shape of the image.

This leads to errors when feeding the images into the tensor into models.

PatWie:
This is not an issue of tensorpack or `imaug.GaussianBlur`. OpenCV already removes the last dimension

````python
import cv2
img = cv2.imread("original.jpg")
grey = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
print grey.shape # is (H, W) instead of (H,W, 1)
````


You should use something like
````python
ds = AugmentImageComponent(ds, [imgaug.MapImage(lambda x: x[..., None])])
````
To get the last axis back.

PatWie:
While I also use grayscale mappings often, I am not sure if it is really a good idea to include this because it depends on the image format rgb or bgr.
You can copy
````python
class Grayscale(ImageAugmentor):
    """
    Convert image to grayscale.
    """

    def __init__(self, keep_dim=False):
        """
        Args:
            keep_dim: return image of shape [H, W, 1] or [H, W]
        """
        self._init(locals())

    def _augment(self, img, _):
        grey = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
        if self.keep_dim:
            grey = grey[..., None]
        return grey
````

into your script.

ppwwyyxx:
Though this is due to opencv, I think it's good to do a reshape inside GaussianBlur to make it less surprising to users (including me).

PatWie:
It is **not** inside GaussianBlur. It is the rgb2gray. At least with opencv 3.1.0.

Skylion007:
Also another point, I am actually not working with grayscale augments, but actually working on a dataset that has grayscale images. Any opencv function call appears to remove the last dimension (not just converting it to grayscale). I am well aware of ways to get the last dimension back and am using them, but I agree with @ppwwyyxx . Unless the operation explicitly changes the shape of an image, it should have the same behavior for multichannel and grayscale images.

PR #131, opened 2017-02-01
Add grayscale mapping

PatWie:
This might solve #129 in a more convenience way

````python
from tensorpack import *

ds = ImageFromFile(["original.jpg"], channel=3)
ds = AugmentImageComponent(ds, [imgaug.Grayscale(keep_dim=True)])
print next(ds.get_data())[0].shape # gives (H, W, 1)

````

### DIFF
--- a/tensorpack/dataflow/imgaug/imgproc.py
+++ b/tensorpack/dataflow/imgaug/imgproc.py
@@ -15,15 +15,20 @@ class Grayscale(ImageAugmentor):
     Convert image to grayscale.
     """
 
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Maybe you do need a rgb/bgr option?

ppwwyyxx:
Thanks! I renamed the class name to CamelCase, and replace `keep_dim` by `keep_dims` used by tensorflow (although it's different from `keepdims` used by numpy)..

PR #132, opened 2017-02-02
LMDB fix

PatWie:
I downloaded ImageNet and converted it into lmdb directly from the tar-files.

Here, I noticed that CaffeLMDB is very slow because its needs to gather the keys by traversing over all 1.2M entries. Here, this would take >35min for ILSVRC2012_img_train.lmdb **during** preparing things.

Now, it is possible to specify the `key_format`, such that there is no overhead anymore and the keys will be generated without any noticeable overhead. Note, the size is already known.

Changes:
- use key_format to generate keys
- remove lmdb_open in reset-state (was open db twice)
- msgpack raised an exception, which is now catched

I can use 
````python
from tensorpack import *

ds = CaffeLMDB('/data/imageNet/ILSVRC2012/lmdb/train', key_format='{:0>8d}')
ds = TestDataSpeed(ds)
ds.start_test()
````
and it will start immediately.

@ppwwyyxx at some point `self.reset_state()` is mixed up in your implementation. I sometimes get `class has no attribute rng` when using MNIST or another dataset.

### DIFF
--- a/tensorpack/dataflow/format.py
+++ b/tensorpack/dataflow/format.py
@@ -59,38 +59,54 @@ class HDF5Data(RNGDataFlow):
 
 class LMDBData(RNGDataFlow):
     """ Read a LMDB database and produce (k,v) pairs """
-    def __init__(self, lmdb_path, shuffle=True):
+    def __init__(self, lmdb_path, shuffle=True, key_format=None):
         """
         Args:
             lmdb_path (str): a directory or a file.
             shuffle (bool): shuffle the keys or not.
+            key_format (str): specify the format of the keys, e.g. `'{:0>8d}'`
         """
         self._lmdb_path = lmdb_path
         self._shuffle = shuffle
-        self.open_lmdb()
+        self.key_format = key_format
+        self.open_lmdb(key_format)
 
-    def open_lmdb(self):
+    def open_lmdb(self, key_format=None):
+        def find_keys(txn, size):
+            logger.warn("Traversing the database to find keys is slow. Your should specify the key_format.")
+            keys = []
+            with timed_operation("Loading LMDB keys ...", log_start=True), \
+                    get_tqdm(total=size) as pbar:
+                for k in self._txn.cursor():
+                    if k[0] != '__keys__':
+                        keys.append(k[0])
+                        pbar.update()
+            return keys
         self._lmdb = lmdb.open(self._lmdb_path,
                                subdir=os.path.isdir(self._lmdb_path),
                                readonly=True, lock=False, readahead=False,
                                map_size=1099511627776 * 2, max_readers=100)
         self._txn = self._lmdb.begin()
         self._size = self._txn.stat()['entries']
+
+        logger.info("Found {} entries in {}".format(self._size, self._lmdb_path))
+
         if self._shuffle:
             # get the list of keys either from __keys__ or by iterating
-            self.keys = loads(self._txn.get('__keys__'))
-            if not self.keys:
-                self.keys = []
-                with timed_operation("Loading LMDB keys ...", log_start=True), \
-                        get_tqdm(total=self._size) as pbar:
-                    for k in self._txn.cursor():
-                        if k[0] != '__keys__':
-                            self.keys.append(k[0])
-                            pbar.update()
+            if key_format is None:
+                try:
+                    self.keys = loads(self._txn.get('__keys__'))
+                except Exception:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ppwwyyxx** [comment]
Thanks for solving this issue! There was an issue about loading keys originally, reset_state() shouldn't need to load all the keys again. This will also slow down the dataflow a lot if you use prefetch.

The original idea was that, lmdb is just a k-v database, and the keys can be any string -- doesn't have to contain numbers. That's why it iterates the db to find the keys.

I like the idea to provide the keys, but the format is assuming the keys are numeric. Maybe you can just add a method in the base lmdb class to set the keys (by giving a list)? Then you can:
```
ds = CaffeLMDB('/data/imageNet/ILSVRC2012/lmdb/train')
ds.set_keys(np.arange(ds.size())
```

**Comment by PatWie** [comment]
The current implementation [can do this](https://github.com/PatWie/tensorpack-fork/blob/lmdb_fix/tensorpack/dataflow/format.py#L106-L107), too.

**Comment by ppwwyyxx** [comment]
Oh I didn't notice that. Could you rename the argument to `keys` and document this as the main feature (and the less-general string formatting as a syntax sugar)?

**Comment by PatWie** [comment]
This is what **I** understand in proper documentation, but probably nobody else.

File: tensorpack/dataflow/format.py [hunk 1]
```diff
@@ -103,6 +118,8 @@ def get_data(self):
                 if k != '__keys__':
                     yield [k, v]
         else:
+            if not hasattr(self, "rng"):
+                self.reset_state()
```
ppwwyyxx: The protocol is that a user always have to call `reset_state()` before `get_data` is first called.

tensorpack/dataflow/format.py
```diff
@@ -59,38 +60,52 @@ def get_data(self):
...
+                    self.keys = loads(self._txn.get('__keys__'))
+                except Exception:
+                    self.keys = find_keys(self._txn, self._size)
+            else:
+                # check if key-format like '{:0>8d}' was given
+                if isinstance(keys, six.string_types):
+                    self.keys = map(lambda x: keys.format(x), list(np.arange(self._size)))
 
     def reset_state(self):
         super(LMDBData, self).reset_state()
```
ppwwyyxx: I wasn't sure if it is safe and efficient to share a lmdb handle among processes so I reopen it again in reset_state(). But the keys don't need to be loaded again.
Removing this line may need some investigation into lmdb.

tensorpack/dataflow/format.py
PatWie: Really? I never did that before. I will look into that again. But this would be Monday.

tensorpack/dataflow/format.py
PatWie: I tried that. Maybe I should also add the converter script such that everybody can test it.

tensorpack/dataflow/format.py
ppwwyyxx: The trainer and TestDataSpeed do that for you. But often you are supposed to call that yourself, although it doesn't always give you exception.

PatWie: Ok. But I want to test it again including the Prefetching(ZMQ) before pushing new code. So it would be Monday at the earliest or Sunday evening.

PatWie:
Ok,
I tested the latest version with:
````python

from tensorpack import *

ds = CaffeLMDB('/data/imageNet/ILSVRC2012/lmdb/train', keys='{:0>8d}')
ds = PrefetchDataZMQ(ds, 4)
# ds = PrefetchData(ds, 4)
ds.reset_state()
for i in ds.get_data():
    img, label = i
    print(img.shape, label)

````
I also prefer to let each process open a file-handle during reset_state as I observed some errors when directly dealing with tar-files.

Works as expected.

ppwwyyxx:
Thanks, this looks good to me now.
One thing I'm thinking about changing (probably later) is: LMDBDataDecoder, CaffeLMDB, etc are  just a mapping on the base LMDBData. If they are implemented with MapData, a change to the base class (like this one) don't have to affect any of them.

PatWie:
Sounds good. One should not forget to add CaffeLMDB as an alias/wrapper, because it is very nice to load Caffe data within one line.

I can do another pull-request with this refactoring, after I spend some time on the testing script to add more test cases and handle/caching the TENSORPACK_DATASET content.

Issue #103, opened 2017-01-15
RNN-cells interface has changed


PatWie:
It seems that the way to import the RNN cell has changed:

````
Traceback (most recent call last):
  File "char-rnn.py", line 192, in <module>
    QueueInputTrainer(config).train()
  File "/home/patwie/git/tensorpack/tensorpack/train/base.py", line 58, in train
    self.setup()
  File "/home/patwie/git/tensorpack/tensorpack/train/base.py", line 130, in setup
    self._setup()
  File "/home/patwie/git/tensorpack/tensorpack/train/feedfree.py", line 101, in _setup
    cost, grads = self._get_cost_and_grad()
  File "/home/patwie/git/tensorpack/tensorpack/train/feedfree.py", line 47, in _get_cost_and_grad
    self.model.build_graph(actual_inputs)
  File "/home/patwie/git/tensorpack/tensorpack/models/model_desc.py", line 108, in build_graph
    self._build_graph(model_inputs)
  File "char-rnn.py", line 72, in _build_graph
    cell = tf.nn.rnn_cell.BasicLSTMCell(num_units=param.rnn_size)
AttributeError: 'module' object has no attribute 'rnn_cell'
````

What do you think about an additional repository https://github.com/tensorpack/data with dummy data, to also test this example using Travis-CI? It seems that in the nearby future they will change a lot of APIs before the final v1.0.

see:
https://github.com/tensorflow/tensorflow/commit/b8101a2b0a7ff1b96999c8e7ad213d1f36b4a4ab

They make it hard to maintain examples. Hopefully, their [release of v1.0](https://github.com/tensorflow/tensorflow/releases/tag/v1.0.0-alpha) will fix that.

ppwwyyxx:
You'll also need to fix tf.nn.rnn, replace it by tf.contrib.rnn.static_rnn.

Maybe it's better to write several dummy training scripts containing lots of features for testing. This may avoid downloading mnist or prepare dummy data for examples. What do you think?

PatWie:
> Maybe it's better to write several dummy training scripts containing lots of features for testing. This may avoid downloading mnist or prepare dummy data for examples. What do you think?

Do you really want to maintain additional examples without any function? Maybe just putting some images (some of image2image produced to not violate copyrights) and a small textfile into a repo and `wget`. The release to run the real examples on very small datasets.

This will require a more flexible testing with teardown to clean up dirs such as `train_logs`

[...]

PR #134, opened 2017-02-04
change to tensorflow v0.12 api for RNN.

cykustcc:
See: https://www.tensorflow.org/api_docs/python/contrib.rnn/this_package_provides_additional_contributed_rnncells_#LSTMBlockCell and https://www.tensorflow.org/api_docs/python/rnn_cell/base_interface_for_all_rnn_cells#RNNCell


PatWie:
There are several other files affected #103

ppwwyyxx:
The code in master now all works for TFv1.0.0rc0 as far as I know. 
Why do you want to change it to work for v0.12?

PR #135, opened 2017-02-05
wasserstein GAN

PatWie:
This is an attempt for a clean TensorFlow implementation of wasserstein-GAN.

- [x] derive child-class from classical GANs
- [x] derive child-class from classical trainer
- [x] collection weights for clipping
- [x] implement symbolic function for clipping
- [x] apply clipping
- [x] produce some results and implement models

### DIFF
--- a/examples/GAN/GAN.py
+++ b/examples/GAN/GAN.py
@@ -1,16 +1,20 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # File: GAN.py
-# Author: Yuxin Wu <ppwwyyxxc@gmail.com>
+# Author: TensorPack contributors
 
 import tensorflow as tf
 import numpy as np
 import time
+import six
 from tensorpack import (FeedfreeTrainerBase, TowerContext,
                         QueueInput, ModelDesc)
 from tensorpack.tfutils.summary import add_moving_summary
 from tensorpack.tfutils.gradproc import apply_grad_processors, CheckGradient
 from tensorpack.dataflow import DataFlow
+from termcolor import colored
+from tensorpack.utils import logger
+import tensorpack.tfutils.symbolic_functions as symbf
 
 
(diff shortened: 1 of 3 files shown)


ppwwyyxx:
I had a wgan implementation locally. Didn't see very surprising results so I didn't intend to push it.
Have you seen anything interesting from your results?

PatWie:
Did you get some results? I think I have all gradients in place but get no results at all.

PatWie:
Ok, MNIST is at least working and gives reasonable results when running all iterations. 
This commit is rebase on last HEAD and contains some refactoring of GAN as well.

ppwwyyxx:
I don't know you're still working on this! Unless WGAN becomes a very common model in the future, I don't think it's necessary to complicate the existing GAN.py. Maybe just a single file like [my gist](https://gist.github.com/ppwwyyxx/f2ee18e08ef140259fd060321680f540) is fine.

There are others factors holding me back otherwise I would have pushed my code earlier already. WGAN reveals some design problems which prevent me from implementing it in easier ways. 
One is that I cannot use anything similar to the existing [VariableClippingOptimizer](https://github.com/tensorflow/tensorflow/blob/33e88f6d7b69f796d4fad452394237448f62b976/tensorflow/contrib/opt/python/training/variable_clipping_optimizer.py) without writing the _setup method myself. I really wish users don't have to touch the trainer for something as simple as clipping the variables.
Another is, when there are >1 ops to run in iterations, running SummaryMovingAverage() together with them may let the graph compute something it shouldn't have (i.e. summary some tensors that were not used by the current op). This doesn't happen in the current implementation, but it could happen with the current design.
I'm thinking about solving the issues (especially the first one) recently, so I could write WGAN easier. Currently this PR and the gist above can be a reference implementation if people want to see it, but I'd like to push it when appropriate changes are made to simplify it.

PatWie:
Aggree on #137

PR #145, opened 2017-02-10
Add helper function/decorator for deprecation warning

PatWie:
This logs warnings about deprecated functions and also shows the
filename as well as the line number.

deprecated

### DIFF
--- a/tensorpack/utils/utils.py
+++ b/tensorpack/utils/utils.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from tqdm import tqdm
 import numpy as np
 import functools
-import logger
+from . import logger
 
 
 __all__ = ['change_env',
@@ -19,6 +19,7 @@ __all__ = ['change_env',
            'get_tqdm',
            'execute_only_once',
            'building_rtfd',
+           'log_deprecated',
            'deprecated'
            ]
 
@@ -65,8 +66,7 @@ def execute_only_once():
     first time and False afterwards.
 
     Returns:
-        bool: whether this is the first time this function gets called from
-            this line of code.
+        bool: whether this is the first time this function gets called from this line of code.
 
     Example:
         .. code-block:: python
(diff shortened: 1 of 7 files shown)


PatWie:
Example:
```python
from tensorpack import *

@deprecated("This is renamed to _get_inputs()", "2017-12-24")
def _get_input_vars(a, b):
    print((a, b))

@deprecated("Use get_reused_placehdrs()")
def get_input_vars(a, b, c):
    print((a, b, c))

_get_input_vars(1, 2)
get_input_vars(3, 4, 5)

deprecated("step_per_epoch")("Use steps_per_epoch", "2017-11-8")
deprecated("config.set_tower")("Set config.tower or config.nr_tower directly.")

class Dummy(object):
    @deprecated("remove this call")
    def test(self, a, b=7):
        pass

def nested():
    d = Dummy()
    d.test(6) 

nested()

deprecated("_get_input_vars() is renamed to _get_inputs().", "2017-04-11")("")

```

tensorpack/utils/utils.py
```diff
@@ -116,3 +119,59 @@ def building_rtfd():
...
+        text (str, optional): information
+        eos (str, optional): end of service date as tuple "YYYY-MM-DD"
+
+    Example:
+        @deprecated("Explanation what to do instead.", "2017-11-4")
+        def foo(...):
+            pass
+
+        deprecated("deprecated_item")("This is an info about an alternative.", "2017-11-4")
+        deprecated("A sentence about a remark.", "2017-11-4")()
```
ppwwyyxx: Shoud it be ("") instead of () ?

PatWie: Yep. But this does not cause the issue. My best guess is that `@functools.wraps(func)` is responsible for that. But I cannot test the python3 version.

PatWie:
Standalone python3 version seems to work:
https://repl.it/FgKc/0

ppwwyyxx:
Now the following two lines are the same:
```python
deprecated("step_per_epoch")("Use steps_per_epoch", "2017-11-8")
deprecated("step_per_epoch", "2017-11-8")("Use steps_per_epoch")
```
What is `fix` and why do you need both `fix` and `eos`?

ppwwyyxx:
It's very weird how I should use the arguments for logging. I don't know if there is a natural way to use one function to work both as a logging function and as a decorator, so I've split the logging to a separate function. Let me know if you have a better solution.

PatWie:
Hi, 
I was also thinking about splitting the deprecation stuff into a function and decorator. The only prerequisite I would say is to allow something like 

```bash
grep -rn . -e "deprecated" --include=*.py
```

There is probably no elegant way to use a decorator as a function itself without messing up with duck typing in python.

PR #164, opened 2017-02-25
allow custom labels for runs

PatWie:
example is in mnist-embeddings


ppwwyyxx:
Adding an extra name like this doesn't help: `dirname + label` can still be an existing directory.

Moreover, since you have a python variable already containing the actual name of the log directory, why still use `auto_set_dir` instead of calling `set_logger_dir` directly? `auto_set_dir` is just to automatically generate a name which "probably is what you want", saving just one line of code.

Issue #170, opened 2017-03-02
add random shifts and 90 deg rotations to imageaugmentor (feature request)


eyaler:
i suggest adding these 2 common augmentations which i could not find:
random horizontal and/or vertical shifts with arbitrary cv2 border behavior 
90 degree rotations or more generally added a "step" param to Rotation (but make sure that 90 deg rotations are exact)

ppwwyyxx:
The second one is just `imgaug.MapImage(lambda x: cv2.transpose(x))` ?

The first one does look like something others may need, but this probably won't come to top of my todo list very soon. Contributions are welcome.

eyaler:
For the second one we want random 90 deg rotations to 0, 90, 180, 270

I will make a PR

PR #177, opened 2017-03-05
added shifts and stepped rotations

eyaler:
https://github.com/ppwwyyxx/tensorpack/issues/170

### DIFF
--- a/tensorpack/dataflow/imgaug/geometry.py
+++ b/tensorpack/dataflow/imgaug/geometry.py
@@ -6,8 +6,33 @@
 from .base import ImageAugmentor
 import math
 import cv2
+import numpy as np
 
-__all__ = ['Rotation', 'RotationAndCropValid']
+__all__ = ['Shift', 'Rotation', 'RotationAndCropValid']
+
+
+class Shift(ImageAugmentor):
+    """ Random horizontal and vertical shifts """
+
+    def __init__(self, horiz_frac=0, vert_frac=0,
+                 border=cv2.BORDER_REPLICATE):
+        """
+        Args:
+            horiz_frac: max abs fraction for horizontal shift
+            vert_frac: max abs fraction for horizontal shift
+            border: cv2 border method
+        """
+        super(Shift, self).__init__()
+        self._init(locals())
+
+    def _get_augment_params(self, img):
+        return np.float32(
+            [[1, 0, np.round(self.horiz_frac * img.shape[1])], [0, 1, np.round(self.vert_frac * img.shape[0])]])
+
+    def _augment(self, img, shift_m):
+        ret = cv2.warpAffine(img, shift_m, img.shape[1::-1],
+                             borderMode=self.border)
+        return ret
 
 
 class Rotation(ImageAugmentor):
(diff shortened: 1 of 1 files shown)


tensorpack/dataflow/imgaug/geometry.py
```diff
@@ -6,30 +6,59 @@
...
+            horiz_frac: max abs fraction for horizontal shift
+            vert_frac: max abs fraction for horizontal shift
+            border: cv2 border method
+        """
+        super(Shift, self).__init__()
+        self._init(locals())
+
+    def _get_augment_params(self, img):
+        return np.float32(
+            [[1, 0, np.round(self.horiz_frac * img.shape[1])], [0, 1, np.round(self.vert_frac * img.shape[0])]])
```
ppwwyyxx: It's not random

tensorpack/dataflow/imgaug/geometry.py
```diff
@@ -43,17 +72,21 @@ class RotationAndCropValid(ImageAugmentor):
         Note that this will produce images of different shapes.
     """
 
-    def __init__(self, max_deg, interp=cv2.INTER_LINEAR):
+    def __init__(self, max_deg, interp=cv2.INTER_LINEAR, step_deg=0):
```
ppwwyyxx: No need to add this feature again. It will be equivalent to Rotation (because there is no need to crop when rotating 90 degrees).

tensorpack/dataflow/imgaug/geometry.py
eyaler: oops.

tensorpack/dataflow/imgaug/geometry.py
eyaler: you could also do steps of 30, 45 etc

ppwwyyxx: OK.. but I can't imagine how that's a useful feature..

eyaler:
i think that rotating (w,h) to (w,h) is indeed the expected behavior, since many algorithms assume a fixed input size. let me know your decision if you want a separate rot90, and if rot90 should have a different behavior. thanks!

ppwwyyxx:
OK, I was thinking about how I would use it, but keeping the size seems useful as well. You can keep it this way.

tensorpack/dataflow/imgaug/geometry.py
ppwwyyxx: If Rotation keeps the size when rotating 90 degree then this augmentor will be different (it doesn't keep the size).

eyaler:
i fixed the shifts to be random. also to make sure 90 deg rotations are pixel exact for odd and even canvas sizes, i made two small changes:

1. use np.round before int in largest_rotated_rect()
2. subtract 0.5 from center param of getRotationMatrix2D

tensorpack/dataflow/imgaug/geometry.py
ppwwyyxx: Why require max_deg==180?

For balanced probability, it's enough to require 2 * max_deg % step_deg == 0.

eyaler: two reasons that relate to the fact that the integer division in not symmetric for positive and negative angles:

1. if max_deg<180 and using step_deg, you could get angles with negative value > max_deg, which breaks user expectation.
2. if max_deg<180 you dont get equal probabilities. say mag_deg=90 and step_deg=90, you get only -90 and 0 angles, missing out on +90

Issue #154, opened 2017-02-19
Refactoring of NCWH broke most GAN examples


PatWie:
Just try to run: InfoGan-mnist or ConditionalGAN-mnist. These should be tested as well #153 
````
patwie@desktop~/git/tensorpack/examples/GAN (master) $ python InfoGAN-mnist.py 
[0219 21:36:42 @logger.py:69] Argv: InfoGAN-mnist.py
[0219 21:36:43 @common.py:101] gen/fc0 input: [None, 74]
[0219 21:36:43 @common.py:109] gen/fc0 output: [None, 1024]
[0219 21:36:43 @common.py:101] gen/fc1 input: [None, 1024]
[0219 21:36:43 @common.py:109] gen/fc1 output: [None, 6272]
[0219 21:36:43 @common.py:101] gen/deconv1 input: [None, 7, 7, 128]
Traceback (most recent call last):
  File "InfoGAN-mnist.py", line 234, in <module>
    GANTrainer(config).train()
  File "/home/patwie/git/tensorpack/tensorpack/train/base.py", line 61, in train
    self.setup()
  File "/home/patwie/git/tensorpack/tensorpack/train/base.py", line 127, in setup
    self._setup()   # subclass will setup the graph
  File "/home/patwie/git/tensorpack/examples/GAN/GAN.py", line 71, in _setup
    self.build_train_tower()
  File "/home/patwie/git/tensorpack/tensorpack/train/feedfree.py", line 40, in build_train_tower
    f()
  File "/home/patwie/git/tensorpack/tensorpack/train/feedfree.py", line 36, in f
    self.model.build_graph(inputs)
  File "/home/patwie/git/tensorpack/tensorpack/models/model_desc.py", line 113, in build_graph
    self._build_graph(model_inputs)
  File "InfoGAN-mnist.py", line 97, in _build_graph
    fake_sample = self.generator(z)
  File "InfoGAN-mnist.py", line 52, in generator
    l = Deconv2D('deconv1', l, [14, 14, 64], 4, 2, nl=BNReLU)
  File "/home/patwie/git/tensorpack/tensorpack/models/common.py", line 104, in wrapped_func
    outputs = func(*args, **actual_args)
  File "/home/patwie/git/tensorpack/tensorpack/models/conv2d.py", line 144, in Deconv2D
    out_channel = out_shape[channel_axis]
IndexError: list index out of range

````

Did you run them?

Issue #153, opened 2017-02-19
Add more tests in travis


PatWie:
The testing script should test some GAN examples as well. Feel free to assign me to this issue.
The tests should use python - unit-tests including tear-down (removing train-log).

PatWie:
See #200

PR #200, opened 2017-03-24
Tests

PatWie:
I started the refactoring of the tests as the previous tests were a bit hacky. Hope this is much easier to maintain. Still not happy with the `setUp` and `teastDown` duplicated code.

This addresses #153 raised in #154. 

- [x] cached MNIST dataset
- [x] testing layers (several examples)
- [x] add dummy images for more complex networks (ideally one should run build ResNet on FakeData)
- [x] remove duplicate codelines

### DIFF
--- a/tests/test_resnet.py
+++ b/tests/test_resnet.py
@@ -3,6 +3,59 @@ import os
 import numpy as np
 import cv2
 import shutil
+import tarfile
+import errno
+from six.moves import urllib
+import sys
+
+
+CAFFE_ILSVRC12_URL = "http://dl.caffe.berkeleyvision.org/caffe_ilsvrc12.tar.gz"
+
+
+# this is duplicated code, but this file should be self-contained
+def mkdir_p(dirname):
+    """ Make a dir recursively, but do nothing if the dir exists
+
+    Args:
+        dirname(str):
+    """
+    assert dirname is not None
+    if dirname == '' or os.path.isdir(dirname):
+        return
+    try:
+        os.makedirs(dirname)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            raise e
+
+
+# this is duplicated code, but this file should be self-contained
+def download(url, dir):
+    """
+    Download URL to a directory. Will figure out the filename automatically
+    from URL.
+    """
+    mkdir_p(dir)
+    fname = url.split('/')[-1]
+    fpath = os.path.join(dir, fname)
+
+    def _progress(count, block_size, total_size):
+        sys.stdout.write('\r>> Downloading %s %.1f%%' %
+                         (fname,
+                             min(float(count * block_size) / total_size,
+                                 1.0) * 100.0))
+        sys.stdout.flush()
+    try:
+        fpath, _ = urllib.request.urlretrieve(url, fpath, reporthook=_progress)
+        statinfo = os.stat(fpath)
+        size = statinfo.st_size
+    except Exception as e:
+        raise e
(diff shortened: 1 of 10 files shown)


PatWie:
There are interesting findings:
- matplotlib is not in requirements.txt. I am not sure if you want to add this.
- the ResNet script relies on "import caffe". This is only for reading the mean-file, right? Do you mind to use a pickle or numpy msgpack file instead?

ppwwyyxx:
+ matplotlib is in opt-requirements, as it's not necessary for most functionailities
+ does it? I think it only imports caffe_pb2.py, which doesn't require caffe to be installed.

PatWie:
@ppwwyyxx Can you have a look at some point. It seems that the ilsvrc-class reads from a different file the paths of the imagenet data. I am trying to fake some image-files an update the "train.txt" and "val.txt".

However, I get
https://travis-ci.org/ppwwyyxx/tensorpack/jobs/215508412#L356

```
 File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/tensorpack/dataflow/dataset/ilsvrc.py", line 184, in get_data
              assert im is not None, fname
          AssertionError: ilsvrc/train/n02481823/n02481823_2939.JPEG
```

This is surprising, as there is not such a line 'ilsvrc/train/n02481823/n02481823_2939.JPEG' in train.txt

ppwwyyxx:
The default dataset directory for `ILSVRCMeta` is `ilsvrc_metadata`, not `ilsvrc`.
But you can pass `meta_dir` argument to `ILSVRC()` as well. It is the full path to the directory.

tests/test_resnet.py
```diff
@@ -0,0 +1,128 @@
...
+import tarfile
+import errno
+from six.moves import urllib
+import sys
+
+
+CAFFE_ILSVRC12_URL = "http://dl.caffe.berkeleyvision.org/caffe_ilsvrc12.tar.gz"
+
+
+# this is duplicated code, but this file should be self-contained
```
ppwwyyxx: why not import stuff from tensorpack? (as another form of testing)

ppwwyyxx:
Btw, resnet example now uses NCHW, which is not supported by TensorFlow on CPU. You might encounter this problem later..

tests/test_resnet.py
PatWie: Sure. I currently thinking about using FakeData by a flag in the ResNet example, which simplifies this testing.

PatWie:
I currently thinking about using FakeData by a flag in the ResNet example, which would simplify this testing.

tests/test_resnet.py
PatWie: the latest version uses the import

PatWie:
After several edits, I was able to convince travis-ci to pass this pull-request. I changed some lines in the resnet example:
- add argument for data_format (for training)
- add argument for faking data

tests/test_resnet.py
```diff
@@ -0,0 +1,43 @@
...
+import os
+import shutil
+import tarfile
+from tensorpack.utils.fs import download, mkdir_p
+
+
+CAFFE_ILSVRC12_URL = "http://dl.caffe.berkeleyvision.org/caffe_ilsvrc12.tar.gz"
+
+
+def fake_ilsvrc12():
```
ppwwyyxx: Is this still needed now that you have --fake?

ppwwyyxx:
Removed fake_ilsvrc12, doesn't seem necessary? 
Thanks very much!

PR #212, opened 2017-03-29
update MinSaver function in saver.py

JimmyCai91:
To copy the current stored tensorflow model requires to save *.index and *.data files, so I rewire the copy lines to save them all.


ppwwyyxx:
Thanks!

Issue #144, opened 2017-02-09
Allow distributed training accross multiple machines


PatWie:
A good reference might be
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/dist_test/python/mnist_replica.py

ahundt:
I suggest looking at http://pipeline.io/ but I haven't used it.

PatWie:
This issue is about re-implementing the mentioned distributed version within TensorPack and not adding more dependencies. Probably the most straight-forward way is something like a `multiple-devices-sync-trainer`.

ahundt:
Sure, I was only trying to provide a helpful reference. :-)

PatWie:
I seems that this requires much many changes. Current I am not sure how the incoming data can be handle in the "tensorpack" way.

But I managed to produce a minimal working example
https://gist.github.com/PatWie/89950d3f8491a9f0d84125dca0945afa

which I actually tested on the following settings:
- machineA: ps + worker, machineB: worker
- machineA: ps + worker + worker

I will try to add a "distributed.py" to the trainer section.

ppwwyyxx:
Thanks for looking at this!
Frankly I haven't looked at tensorflow distributed training in depth yet. Because I found now it's still hard to feed 8 GPUs on one machine, let alone multiple machines. That's why I'm working on data recently.

btw, TF say they'll publish a "good" example code of distributed training very soon, with the use of GPU StageArea. It'll include the best practice so I guess we can refer to that (rather than finding it out by ourselves) when it come out.

ahundt:
Cool do you have a link to where they reference this?

Also, one advantage of distributed training is you can amortize io to ease feeding the GPUs. Also, as an example in my case I have access to many 2 GPU machines and zero 8 GPU machines.

PatWie:
I looks like there was an annoucement at the devSummit. I rarely see machines that handle 8GPUs and each machine could locally read from the SSD. So the bottleneck would be at some other point.

ppwwyyxx:
Official code and benchmarks were released:
https://www.tensorflow.org/performance/performance_models
https://www.tensorflow.org/performance/benchmarks

I'll study it when I have time and see how much we can integrate them.

PatWie:
Why did you remove the `examples/distributed.py`?

ppwwyyxx:
Just feel mnist is too boring..

PatWie:
[...]

Issue #193, opened 2017-03-17
multi gpu + multi tower on slurm cluster


ahundt:
I'm trying to integrate a couple of different components and was hoping to seek advice. TensorPack looks like it has some good [multigpu scripts](https://github.com/ppwwyyxx/tensorpack/blob/master/tensorpack/train/multigpu.py) I'd like to employ for both image labeling and semantic segmentation training.

I have the following:

 - [tensorflow_slurm_manager](https://github.com/jhollowayj/tensorflow_slurm_manager) to get the [tf.train.ClusterSpec](https://www.tensorflow.org/api_docs/python/tf/train/ClusterSpec) once things have started up
 - [slurm cluster with gpus](https://slurm.schedmd.com/)
     - [relevant discussion](https://github.com/jhollowayj/tensorflow_slurm_manager/issues/1) and [exampe](https://github.com/jhollowayj/distributed_tensorflow/blob/slurm_distributed/main.py#L63-71)
 - a couple keras models to train, being discussed in https://github.com/ppwwyyxx/tensorpack/issues/160

A slurm script is just some comments at the top of a shell script that specifies the number of machines, gpus etc and gets lists of what was allocated then runs the programs you wanted later in the shell script.

Basically, examples and usage of AsyncMultiGPUTrainer seem a bit opaque at the moment, though I do see one in the [A3C code](https://github.com/ppwwyyxx/tensorpack/blob/master/examples/A3C-Gym/train-atari.py#L273), and not much that works with the standard TensorFlow ClusterSpec or a multi tower example. 

With all that I have a few questions:

 - Can a multi tower example be created, and if it makes sense to integrate it with a `ClusterSpec`?
 - Will there be any limitations I should be aware of with the TensorPack APIs when training with segmentation data?

ppwwyyxx:
Do you just want multiGPU? Distributed training are not supported at the moment.

If just multiGPU, is slurm anywhere different from a normal machine with multiple GPUs? I know nothing about slurm so I'm not sure what's the issue.

examples/HED is a segmentation example. Nothing really different from other tasks.

ahundt:
[...]

PR #223, opened 2017-04-08
Distributed Trainer

PatWie:
Distributed means several training processes (even accross multiple machines).

Not sure if this is possible without rewriting large portions of your code. At least it is an attempt for #144. I mostly focus on transforming the [MWE](https://gist.github.com/PatWie/89950d3f8491a9f0d84125dca0945afa) into a trainer-class.

There are still several things to do:

- [x] Where to place the  SyncReplicasOptimizer? ModelDesc?
- [ ] a `prepare_or_wait_for_session` sessinit class.
- [x] Where does the data comes from? ZeroMQ? Currently, it is multiple processes for each worker.
- [x] I though there was a global_step tensor function, which seems to be required in `tf.train.Supervisor`

I currently just try to use the SyncMultiGPUTrainer before make it more complex. I cannot estimate how complex this feature will be. 
@ppwwyyxx you probably know if the distributed session management currently fits sessinit.py and sesscreate.py. This is the part I never read :wink:

### DIFF
--- a/examples/mnist-distributed.py
+++ b/examples/mnist-distributed.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# File: mnist-convnet.py
+# Author: Yuxin Wu <ppwwyyxxc@gmail.com>
+
+import numpy as np
+import os
+import sys
+import argparse
+
+"""
+python mnist-distributed.py --task_id 0 --job ps --gpu 0
+python mnist-distributed.py --task_id 0 --job worker --gpu 0
+python mnist-distributed.py --task_id 1 --job worker --gpu 0
+
+
+
+"""
+
+# Just import everything into current namespace
+from tensorpack import *
+import tensorflow as tf
+import tensorflow.contrib.slim as slim
+
+IMAGE_SIZE = 28
+
+
+class Model(ModelDesc):
+    def _get_inputs(self):
+        return [InputDesc(tf.float32, (None, IMAGE_SIZE, IMAGE_SIZE), 'input'),
+                InputDesc(tf.int32, (None,), 'label')]
+
+    def _build_graph(self, inputs):
+
+        image, label = inputs
+        image = tf.expand_dims(image, 3)
+        image = image * 2 - 1   # center the pixels values at zero
+
+        with argscope(Conv2D, kernel_shape=3, nl=tf.nn.relu, out_channel=32):
+            logits = (LinearWrap(image)  # the starting brace is only for line-breaking
+                      .Conv2D('conv0')
+                      .MaxPooling('pool0', 2)
+                      .Conv2D('conv1')
+                      .Conv2D('conv2')
+                      .MaxPooling('pool1', 2)
+                      .Conv2D('conv3')
+                      .FullyConnected('fc0', 512, nl=tf.nn.relu)
+                      .Dropout('dropout', 0.5)
+                      .FullyConnected('fc1', out_dim=10, nl=tf.identity)())
+
+        prob = tf.nn.softmax(logits, name='prob')   # a Bx10 with probabilities
+
+        cost = tf.nn.sparse_softmax_cross_entropy_with_logits(logits=logits, labels=label)
+        cost = tf.reduce_mean(cost, name='cross_entropy_loss')  # the average cross-entropy loss
+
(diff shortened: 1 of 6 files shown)


ahundt:
Coult it be worth considering providing the clusterspec as a parameter?

This is because depending on the distributed system the most sensible manner to initialize the clusterspec will vary, as mentioned in https://github.com/ppwwyyxx/tensorpack/issues/193

PatWie:
As far as I know, it is currently not possible to specify the GPUs for each worker. That's why I used a dict instead.

My current version (I haven't pushed yet) just creates a ClusterTrainer relying on clusterSpec and uses all getting all GPU from CUDA_VISIBLE_DEVICES like the default behavior of TF. But the code is currently not complete. I will post any updates this week if I have time to code.

PatWie:
Now, 
a few updates and findings on this:

- the global_step tensor has to be placed on a [specific device](https://github.com/tensorflow/tensorflow/blob/17c47804b86e340203d451125a721310033710f1/tensorflow/python/training/supervisor.py#L879-L883)
- sometimes TF needs the `global_step` op in the optimizer call. It is required to prevent a stale optimizer.
- there is an complex interplay between the MonitoredSession and the `tf.train.Supervisor` and I am not willing to re-implement most parts of the `Supervisor` class.
- it requires a lot of additional changes in terms of *where* to log and *when* to log or when to write a summary. It seems that the `tf.train.Supervisor` handles this by [knowing whether](https://github.com/tensorflow/tensorflow/blob/17c47804b86e340203d451125a721310033710f1/tensorflow/python/training/supervisor.py#L657-L667) the worker is a chief-worker or not.

For the current version I set the correct ip addresses in:
```python
MACHINE_Q = 'xxx.x.xx.xxx'  # noqa
MACHINE_G = 'xxx.x.xx.xxx'  # noqa
```

and then fire up the example by
```
python mnist-distributed.py --task_id 0 --job ps --gpu 0 # on MACHINE_Q
python mnist-distributed.py --task_id 0 --job worker --gpu 0 # on MACHINE_Q
python mnist-distributed.py --task_id 1 --job worker --gpu 0 # on MACHINE_G
```

The graph is building fine on all machines.  Though I currently get an issue with the [monitored session](https://github.com/tensorflow/tensorflow/blob/52dcb2590bb9274262656c958c105cb5e5cc1300/tensorflow/python/training/monitored_session.py#L428).
```
AttributeError: 'DistributedTrainer' object has no attribute '_monitored_sess'
2017-04-20 11:04:53.496895: E tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:647] failed to record completion event; therefore, failed to create inter-stream dependency
2017-04-20 11:04:53.496958: I tensorflow/stream_executor/stream.cc:4125] stream 0x5c544a0 did not memcpy device-to-host; source: 0x330a3c0800
2017-04-20 11:04:53.496972: E tensorflow/stream_executor/stream.cc:289] Error recording event in stream: error recording CUDA event on stream 0x5c54570: CUDA_ERROR_DEINITIALIZED; not marking stream as bad, as the Event object may be at fault. Monitor for further errors.
2017-04-20 11:04:53.496989: E tensorflow/stream_executor/cuda/cuda_event.cc:49] Error polling for event status: failed to query event: CUDA_ERROR_DEINITIALIZED
2017-04-20 11:04:53.497000: F tensorflow/core/common_runtime/gpu/gpu_event_mgr.cc:203] Unexpected Event status: 1
Aborted
```

It seems that the `sessioncreate.py` in this repo already should distinguish between chief-workers and non-chief-workers. The relevant reference is in https://github.com/tensorflow/tensorflow/blob/52dcb2590bb9274262656c958c105cb5e5cc1300/tensorflow/python/training/monitored_session.py#L343

I

PatWie:
I messed up the last commit and will rebase again.

ppwwyyxx:
Thanks very much for trying out! I'll play with your code soon.

PatWie:
Their tutorial describes a distributed **async** training. But this pull request deals with a **sync** training procedure, which I found more intuitively and less hacky. I have no experience which version is faster (async/sync) or better. But I guess this depends on the actual problem.

ppwwyyxx:
In sync training, the optimizer needs to maintain the global_step by itself, so I shouldn't use a callback to maintain it. That's a problem.

PatWie:
closed in favor of the distributed branch. Nice work @ppwwyyxx !!

Issue #207, opened 2017-03-27
Duplicating Datapoints in DataFlow


msbauer:
Hello,

I would like to duplicate datapoints in a DataFlow, so I wrote a ProxyDataFlow:
```
class RepeatedDataPoint(ProxyDataFlow):

    """ Take data points from another DataFlow and produce them a certain number of times
    dp1, ..., dp1, dp2, ..., dp2, ...
    """

    def __init__(self, ds, nr):
        """
        Args:
            ds (DataFlow): input DataFlow
            nr (int): number of times to repeat each datapoint.
        """
        self.nr = nr
        super(RepeatedDataPoint, self).__init__(ds)

    def size(self):
        """
        Raises:
            :class:`ValueError` when nr == -1.
        """
        return self.ds.size() * self.nr

    def get_data(self):
        for dp in self.ds.get_data():
            for _ in range(self.nr):
                yield dp
```

However, it does not seem to duplicate the datapoint at this point but also including augmentors applied afterwards:
```
ds = ... # shape (2000, 3000, 1)
ds = RepeatedDataPoint(ds, nr=5)
ds = AugmentImageComponent(ds, [imgaug.RandomCrop((256, 256))])
ds = PrintData(ds, num=5)
```
yields
```
datapoint 0<5 with 1 components consists of
   dp 0: is ndarray of shape (256, 256, 1) with range [-0.8151, 1.3173]
datapoint 1<5 with 1 components consists of
   dp 0: is ndarray of shape (256, 256, 1) with range [-0.8151, 1.3173]
datapoint 2<5 with 1 components consists of
   dp 0: is ndarray of shape (256, 256, 1) with range [-0.8151, 1.3173]
datapoint 3<5 with 1 components consists of
   dp 0: is ndarray of shape (256, 256, 1) with range [-0.8151, 1.3173]
datapoint 4<5 with 1 components consists of
   dp 0: is ndarray of shape (256, 256, 1) with range [-0.8151, 1.3173]
```
i.e. it always chooses the same patch because diffh==0 and diffw==0 after the first augmentation.

I presume I am doing something wrong. If so, please let me know what it is. Alternatively: Is there a better way to duplicate datapoints (I read large high-resolution files from disk. From each of them I want to extract many random crops. However, I don't want to re-read them every time.

Thanks
Matthias

PatWie:
You need to backup the current point like
```
    for dp in d:
        dp_current = dp
        for _ in range(5):
            yield dp_current
[...]

PR #227, opened 2017-04-10
solution from issue 207

PatWie:
see #207


ppwwyyxx:
The augmentor now would copy by default. So that issue was fixed

PatWie:
This is not the copy parts. These changes are about the RepeatedDataPoint.

ppwwyyxx:
If no dataflow makes modifications by default, then no dataflow needs to make an explicit copy for potential modification.

PatWie:
Removed the copy-function. But this data-mapping is quite common.

PR #253, opened 2017-05-05
Augment Tensors from Layers (fix #228)

### LINKED ISSUE #228: [RFC] Access layer attributes
Continue on #204 

Problem: No easy access to weights/bias/trainable_variables, etc, of Conv2D

Several possible designs:
1. 
```python
x = Conv2D(...)
w = x.weights
```
This is a hack on `tf.Tensor`. Not sure if it'll break in the future.

2. 
```python
x = Conv2D(...)
w = Conv2D.weights
```
This is used by [chi](https://github.com/rmst/chi/).
Things start to get hard when you need to use Conv2D multiple times.

3. Maintain a global registry of (tensor -> attributes)
```python
x = Conv2D(...)
w = LayerAttributes(x).weights
```

PatWie:
After writing `x = Conv2D(...)` one can now access the used variables
by `w = x.variables.W` and `b = x.variables.b` if they exists.

This is much better than writing something like:
```
w = x.name.replace('/output:0', "/W:0").replace('tower0/', '')
w = tf.get_default_graph().get_tensor_by_name(w)
```
which would also break the InferenceRunner.

Now, I also rely on this feature. And I think this the most straightforward way to access the variables. I did a grep over the *.py files in the TensorFlow repo and so far found no collision.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ install:
 
 before_script:
 - flake8 --version
+- protoc --version
 - python -c "import cv2; print('OpenCV '+ cv2.__version__)"
 - python -c "import tensorflow as tf; print('TensorFlow '+ tf.__version__)"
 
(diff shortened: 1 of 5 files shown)


ppwwyyxx:
I like this interface. Not too hacky and convenient.

PR #258, opened 2017-05-08
Simplify simple examples

PatWie:
the mnist-convnet.py examples was overloaded. This commit splits the
file into the most minimal mnist-convnet and mnist-tfslim. Further
it adds an example about the visualization of filters which
illustrates the usage of `VariableHolder`.

In addition, this also adds a boilerplate template for projects,
which might facilitate the process of starting a new project.

### DIFF
--- a/examples/boilerplate.py
+++ b/examples/boilerplate.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Author: Your Name <your@email.com>
+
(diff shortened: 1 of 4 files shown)


PatWie:
A comment:
I think it is hard to parse the current mnist-convnet example when starting working with tensorpack. Further, I always start by trimming the MNIST example, when I start a new project. I think a well-maintained boilerplate template without deprecation warnings is a nice start point.

PR #259, opened 2017-05-08
fix #236

### LINKED ISSUE #236: Feature request: Support for depth-wise and separable convolutions
TF has APIs for this but if tensorpack abstracts these APIs out it will be helpful.

PatWie:


PatWie:
It might be worth to refactor this into a class with `__call__` to reuse code.

ppwwyyxx:
I feel very confused about what it really does when these convolutions are mixed in one function, especially when the concept of "separable" and "depthwise" already kind of overlap.

I hope we can just provide individual functions so that they have their corresponding counterparts in `tf.nn.*` or `tf.layers.*`, with the same set of arguments.

ppwwyyxx:
In the future we can hopefully be free from maintaining all kinds of layers and just copy or wrap `tf.layers` or other well-maintained wrappers. `tf.layers` have the attribute accessor as well (`.variables`, etc) so this could be possible. The only inconsistency now is the naming (W vs weights, out_channel vs filters, etc).

PatWie:
Just wrapping tf.layers will be a breaking change as you will never be able to load old models again without a lot of code in the ModelLoader. Further, some layers like BatchNorm in your lib have some extra versions (local stats) which are helpful.
So, we need 
- mnist-prettytensor.py
- mnist-sonnet.py
- mnist-tf.layers.py
- mnist-sugartensor.py

? :-)

Maybe, providing the common layers is ok. So I closed this pull-request.

PR #260, opened 2017-05-09
fix bug in boilerplate

PatWie:
Change FakeData to support different data types and ranges of values. Motivation is to enable ci-tests for the boilerplate template.

### DIFF
--- a/tensorpack/dataflow/raw.py
+++ b/tensorpack/dataflow/raw.py
@@ -40,13 +40,13 @@ class FakeData(RNGDataFlow):
             for _ in range(self._size):
                 val = []
                 for k in range(len(self.shapes)):
-                    v = self.rng.rand(*self.shapes[k]) * self.domain[k][1] + self.domain[k][0]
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
Why will the domain be useful?

PatWie:
You need the `domain` for fake-labels of type uint8.

tensorpack/dataflow/raw.py
```diff
@@ -14,32 +15,41 @@
...
     def size(self):
         return self._size
 
     def get_data(self):
         if self.random:
             for _ in range(self._size):
-                yield [self.rng.rand(*k).astype(self.dtype) for k in self.shapes]
+                val = []
+                for k in range(len(self.shapes)):
+                    v = self.rng.rand(*self.shapes[k]) * self.domain[k][1] + self.domain[k][0]
```
ppwwyyxx: Domain means (min, range) ? Would it be simpler to use (min, max)?

PatWie:
You are right!

PR #267, opened 2017-05-11
Clarify LMDB-dumps and MinSaver in Documentation

PatWie:
The user might be familiar with creating a ds-stream and wants to
dump the compressed image to disk as an lmdb file. This adds this to
the documentation but gives a warning about the write-performance.

In addition, there is a small amount of  documentation about the
Min-Saver and ScalarStats-runner, which is slightly now described in
more detail.

### DIFF
--- a/docs/tutorial/efficient-dataflow.md
+++ b/docs/tutorial/efficient-dataflow.md
@@ -123,14 +123,25 @@ ds1 = PrefetchDataZMQ(ds0, nr_proc=1)
 dftools.dump_dataflow_to_lmdb(ds1, '/path/to/ILSVRC-train.lmdb')
 ```
 The above script builds a DataFlow which produces jpeg-encoded ImageNet data.
-We store the jpeg string as a numpy array because the function `cv2.imdecode` expect it later.
+We store the jpeg string as a numpy array to save disk space compared an uncompressed collection of images (like CaffeLMDB) and to save memory bandwidth during reading them later.
 We use 1 prefetch process to speed up. If `nr_proc>1`, `ds1` will take data
-from several forks of `ds0` and will not be identical to `ds0` any more.
+from several forks of `ds0` and will not be identical to `ds0` any more. It will generate a database file of 140G and takes roughly 25 minutes on a modern system with an SSD.
 
-It will generate a database file of 140G. We build a DataFlow to read the LMDB file sequentially:
+You can serialize your custom dataflow into such a LMDB file by
+
+```python
+ds0 = dataset.ILSVRC12('/path/to/ILSVRC', 'train', shuffle=False)  # or some custom streams
+ds0 = ImageEncode(ds0, mode='.jpg', index=0)
+ds1 = PrefetchDataZMQ(ds0, nr_proc=1)
+dftools.dump_dataflow_to_lmdb(ds1, '/path/to/ILSVRC-train.lmdb')
 ```
+
+While this offers more flexibility, it is much slower during writing (980 it/s vs. 67 it/s) because OpenCV has to decode (`dataset.ILSVRC12`) first and later encode (`ImageEncode`) the image. But in some cases, like a pre-processed validation set or randomly extracted frames from a large collection of videos, it is convenient to compress extracted images by `ImageEncode`.
+
+We build a DataFlow to read the LMDB file sequentially:
+```python
 from tensorpack import *
-ds = LMDBData('/path/to/ILSVRC-train.lmdb', shuffle=False)
+ds = LMDBDataPoint('/path/to/ILSVRC-train.lmdb', shuffle=False)
 ds = BatchData(ds, 256, use_list=True)
 TestDataSpeed(ds).start_test()
 ```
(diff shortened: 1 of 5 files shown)


PatWie:
@ppwwyyxx Are you really sure, that the reading speed is just 130 it/s ? I get 400 it/s on local SSD.

docs/tutorial/efficient-dataflow.md
```diff
@@ -157,19 +168,19 @@ Then we add necessary transformations:
 .. code-block:: python
     :emphasize-lines: 3-5
 
-    ds = LMDBData(db, shuffle=False)
+    ds = LMDBData(db, shuffle=False) # produces key-value pairs
     ds = LocallyShuffleData(ds, 50000)
-    ds = LMDBDataPoint(ds)
-    ds = MapDataComponent(ds, lambda x: cv2.imdecode(x, cv2.IMREAD_COLOR), 0)
+    ds = LMDBDataPoint(ds)  # converts 'ds' to  pairs (encoded_image, label)
+    ds = ImageDecode(ds, mode='.jpg', index=0)
```
ppwwyyxx: Can we just keep the Map? It's a one-liner anyway, and user can see and choose what decoder they want (I know faster decoder than opencv). The wrappers also lose the options available in imdecode/imencode (IMREAD mode, image quality, etc).

ppwwyyxx:
Probably RAM on my test machines wasn't very good. I just ran a hdparm benchmark on a new machine and see 11G/s cached read speed. This seems to be close to the speed you get.

ppwwyyxx:
No. 130it/s is after batching to 256.

docs/tutorial/efficient-dataflow.md
PatWie: `SelectComponent`, `ColorSpace`, `Grayscale`, `ToUint8` , `ToFloat32`, `FixedCrop`, `Clip`, `Identity`, `JpegNoise`, `Resize` are also one-liner.

ppwwyyxx: Yeah I think FixedCrop is especially unnecessary. When I see the function I'll be wondering whether it crops to max+1 or max.

tensorpack/dataflow/image.py
```diff
@@ -11,7 +11,39 @@
...
+class ImageDecode(MapDataComponent):
+    """ Decode image as a string. """
+    def __init__(self, ds, index=0, mode=cv2.IMREAD_COLOR):
+        """
+        Args:
+            ds (DataFlow): input DataFlow.
+            index (int): the index of the component to be decoded.
+        """
+        def func(im_data):
+            return cv2.imdecode(np.asarray(bytearray(im_data), dtype=np.uint8), mode)
```
ppwwyyxx: Would bytearray introduce a copy when input is already a uint8 nparray?..

docs/tutorial/efficient-dataflow.md
ppwwyyxx: Also, if it is provided as a dataflow called `ImageDecode`, people would want to use it with one datatype or another (at least bytes and nparray are common). Similarly for encode, we cannot assume users want an nparray in the end, bytes can be useful as well.
 Although they seem so, these two functions are not so well-defined like `Clip`,`ToUint8`,`Resize` which you don't need to read docs to know what they do.
I don't mind removing `SelectComponent`, actually I would hope to let users get familiar with `MapData/MapImage` so they won't expect all simple functionalities built into tensorpack.

tensorpack/dataflow/image.py
PatWie: It does

PatWie:
I would still vote for these atomic on-liners `ImageEncode`, `ImageDecode`  as they are pretty common in any good LMDB based training procedure. Please note, that the code from the current documentation does not work:

```python
ds = LMDBData(db, shuffle=False)
ds = LocallyShuffleData(ds, 50000)
ds = LMDBDataPoint(ds)
```
In `LMDBDataPoint` there is a `isinstance(args[0], LMDBData)` which is obviously `False` if there is a `LocallyShuffleData` between.

Maybe a renaming from
`LMDBDataPoint` to `MsgPackDecoder` or `LMDBdumpDeoder` is also more intuitiv.

tensorpack/dataflow/format.py
```diff
@@ -160,16 +160,15 @@ class LMDBDataPoint(MapData):
     """ Read a LMDB file and produce deserialized values.
         This can work with :func:`tensorpack.dataflow.dftools.dump_dataflow_to_lmdb`. """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, ds):
         """
         Args:
-            args, kwargs: Same as in :class:`LMDBData`.
-        """
-        if isinstance(args[0], LMDBData):
```
PatWie: This is almost never True.

ppwwyyxx: Right.

ppwwyyxx:
I agree they are common and I'm only concerned with the type confusion. "Encode" always means "convert to bytes" except in opencv. In fact, encoding to nparray is only useful when you decode with opencv. 
When we add distributed training, we probably have to use jpeg-encoded __bytes__ in a TFRecord and decode with TF, to reach maximum speed. At least that's what google is doing. In that case we'll need encode/decode to handle bytes.
I don't mind adding it if it could at least support bytes, but it will just look too complicated (both the docs and the code) given that it could be done in one line. Originally I used bytes in the tutorial and I only added nparray conversion for speed (and not even sure if it really helps).

I agree on the rename. But `MsgPackDecoder` is not very good because it decodes `dp[1]`, which is an unexpected behavior from the name. Decoding from `dp[1]` is LMDB-specific behavior (just like decoding nparray is cv-specific behavior) so we better keep LMDB in the name.

PatWie:
At least there should be a convenient usage:
- get data
- encode
- dump

and then
- load
- decode

What about a combined `Serializer` class which can return properties like a data stream for reading and writing, like

```
ds = ...
wrapper = MyFileTypeWrapper(file='/tmp/test.lmdb', format='lmdb')
wrapper.serialize(ds)
```

and then

```
wrapper = MyFileTypeWrapper(file='/tmp/test.lmdb', format='lmdb')
wrapper.deserialize(ds) # a combined version of image-encode, decode and loads and dump
```

ppwwyyxx:
I didn't get why this is needed. What's the benefit of using a new API serialize/deserialize compared to the original APIs `dump_to_lmdb` and `LMDBDataPoint`? You need to write decode/encode somewhere anyway.

PatWie:
Closing for now. But you should remove the lines:
https://github.com/ppwwyyxx/tensorpack/blob/master/tensorpack/dataflow/format.py#L168-L171
as we discussed here:
https://github.com/ppwwyyxx/tensorpack/pull/267/files#diff-4a08f729b0314e28cdde668081b6ce22

PR #269, opened 2017-05-11
Fix incorrect slicing in expreplay when curr_size != max_size.

rhofour:
Previously, this would make s1 extend to contain all of the
uninitialized elements.


ppwwyyxx:
Slice is only useful when the memory buffer is full

rhofour:
Ah, looking closer at how sample is called I see that's the case.

Looks like the real bug is in my prioritized experience replay implementation which handles this case incorrectly.

Issue #239, opened 2017-04-25
Pimp the tf.image to the tensorboard on validset in OnlineExport(callback)


tmquan:
Hi tensorpack, 

I would like to ask this issue:
I am using image2image.py and training with 2 datasets ds_train and ds_valid declared in 
```python
def get_config():
    logger.auto_set_dir()
    # dataset = get_data()
    ds_train, ds_valid = get_data()
    ds_train.reset_state()
    ds_valid.reset_state() 
    #print ds_train.size()
    return TrainConfig(
        dataflow=ds_train,
        callbacks=[
            PeriodicTrigger(ModelSaver(), every_k_epochs=100),
            ScheduledHyperParamSetter('learning_rate', [(200, 1e-4)]),
            PeriodicTrigger(InferenceRunner(ds_valid, [ScalarStats('L1_loss')]), every_k_epochs=1),
        ],
        model=Model(),
        steps_per_epoch=ds_train.size(),
        max_epoch=4000,
    )
```

The OnlineExport callback (http://tensorpack.readthedocs.io/en/latest/casestudies/colorize.html#callbacks) allows me to generate the input and predicted outputs as well as the related loss values but it doesn't show up similar tf.image in the tensorboard. 

Is there anyway to pimp the results of prediction on ds_valid to the tensorboard via this callback?

Best regards,

ppwwyyxx:
There is a plan about it but not ongoing.
Basically the goal is to let the [monitor](https://github.com/ppwwyyxx/tensorpack/blob/master/tensorpack/callbacks/monitor.py) support data types other than scalar.

Skylion007:
Also a request would be if you could allow the callback to say save images of the validation set in a folder. That would be extremely useful is one of the main issues I have with the current repository. Is there a way to hack inference runner to do this currently?

ppwwyyxx:
If it is just a fixed validation set, why do you need to do it during training instead of before training?

And you can do that by writing an Inferencer. No need to hack anything.

PatWie:
The new tensorboard has a slider below each image-summary. If `tf.summary.image` is called on the same input, one could inspect the result directly in the browser from different epochs.

ppwwyyxx:
You can now use `self.trainer.monitors.put_image(name, array)` to put a numpy array to tensorboard.

PR #272, opened 2017-05-16
Dynamic Filter network

PatWie:
Dynamic Filter Networks <https://arxiv.org/abs/1605.09673>
by Bert De Brabandere et al.

This also demonstrates `monitors.put_image` from #239, which gives a nice slider for the `filter_export` summary in the newest TF version.


ppwwyyxx:
Looks like this is section4.1 of the paper. 
Were you able to see the network learning expected filters? If it's successful it would be a good example to add.

PatWie:
if you run this network, you can use tensorboard and will see these filters after a while. Tensorboard will show (OnlineTensorboardCallback) both the expected and learned filters.

ppwwyyxx:
Since they did open source their code https://github.com/dbbert/dfn/blob/master/experiment_steerableFilter_tensorflow.ipynb, I guess it would be better to follow the network & hyperparam rather than making things up.
Plus they seem to have a more efficient dynamic conv implementation.

PatWie:
I changed the conv2d to fullyconnected layers and get a faster convergence:

tensorboard_dfn

PatWie:
I don't think, that their "dynamic conv implementation" is more efficient (in terms of speed *and* memory). Rather than  using their explicit im2col, this implementation uses a highly optimized call of TF. 
Further, TensorFlow uses the cuda streams, so this would also hide the latency from data transfers between kernel calls from the `[... zip(...)]`.

ppwwyyxx:
Do you know why they get to a very small loss with only 490 steps (and their loss is summed over the image and averaged over batch). Maybe because their dataset is easier?
Anyway the current result you showed is pretty good.

PatWie:
I have collected a large image data set, where I simply rejected images without any prominent edges. Using this custom dataset helps a lot to converge faster.
So, I would guess it depends heavily on the training data, e.g. try MNIST.

examples/DynamicFilterNetwork/steering-filter.py
```diff
@@ -0,0 +1,286 @@
...
+
+        Returns:
+            learned filter as [B, k, k]
+        """
+        with argscope(LeakyReLU, alpha=0.2), \
+                argscope(FullyConnected, nl=LeakyReLU):
+            net = FullyConnected('fc1', theta, 64)
+            net = FullyConnected('fc2', net, 128)
+
+        pred_filter = FullyConnected('conv3', net, kernel_shape ** 2, nl=tf.identity)
```
ppwwyyxx: Probably 'conv3' isn't the desired name?

examples/DynamicFilterNetwork/steering-filter.py
```diff
@@ -0,0 +1,286 @@
...
+        gt_image = tf.reshape(gt_image, [BATCH, SHAPE, SHAPE, 1])
+
+        pred_filter = self._parameter_net(theta)
+        pred_image = dynamic_conv_filter('dfn', image, pred_filter, 1, kernel_size)
+
+        with tf.name_scope('visualization'):
+            def n(x):
+                x = x - tf.reduce_min(x)
+                x = x / tf.reduce_max(x)
+                return x
```
ppwwyyxx: tf.summary.image will do some kind of normalization when input is float. Maybe this is not necessary?

examples/DynamicFilterNetwork/steering-filter.py
```diff
@@ -0,0 +1,286 @@
...
+    rsl = tf.concat(rsl, axis=0, name='output')
+    return rsl
+
+
+class OnlineTensorboardExport(Callback):
+    """Show learned filters for specific thetas in TensorBoard.
+    """
+    def __init__(self):
+        # generate 32 filters (8 different, 4 times repeated)
+        self.theta = np.array(range(8) * 4) / 8. * 2 * np.pi
```
ppwwyyxx: `range(8)*4` is not python3 compatible.

examples/DynamicFilterNetwork/steering-filter.py
```diff
@@ -0,0 +1,286 @@
...
+            theta = self.rng.uniform(0, 2 * np.pi)
+            filtered_image, used_filter = ThetaImages.filter_with_theta(image[0], theta)
+            yield [theta, image[0], filtered_image, used_filter]
+
+
+def get_data():
+    # probably not the best dataset
+    ds = dataset.BSDS500('train', shuffle=True)
+    # rgb2gray
+    ds = MapData(ds, lambda dp: [np.dot(dp[0], [0.299, 0.587, 0.114])[:, :]])
```
ppwwyyxx: You added `Grayscale ` augmentor but is not using it ..

examples/DynamicFilterNetwork/steering-filter.py
PatWie: Why do you not just give my creativity some freedom? ;-)

PatWie:
I changed these lines and also partly tested python3 compatibility.

PR #275, opened 2017-05-18
Corrected tensorflow module path for StagingArea

Skylion007:
Bugfix for StagingArea. It appears as though Tensorflow is not adding it in the old contrib location. Should be backwards compatible.

### DIFF
--- a/tensorpack/train/input_source.py
+++ b/tensorpack/train/input_source.py
@@ -5,7 +5,7 @@
 
 import tensorflow as tf
 try:
-    from tensorflow.contrib.staging import StagingArea
+    from tensorflow.python.ops.data_flow_ops import StagingArea
 except ImportError:
     pass
 


ppwwyyxx:
StagingArea will be added to contrib, if I'm not mistaken. It's now in contrib in nightly.

tensorpack/train/input_source.py
```diff
@@ -5,7 +5,7 @@
 
 import tensorflow as tf
 try:
-    from tensorflow.contrib.staging import StagingArea
+    from tensorflow.python.ops.data_flow_ops import StagingArea
 except ImportError:
     pass
```
PatWie: I really enjoy this empty catch block. I know somebody who had some issues with this line.

Skylion007:
@ppwwyyxx So what? This does the same thing as it being added to nightly. Check out the file they add to [contrib](https://github.com/tensorflow/tensorflow/blob/ffd1ed2df78bdc719abe6180db896b26d04d1b70/tensorflow/contrib/staging/__init__.py).

Skylion007:
To underscore my point, here is the entirety of contrib staging module minus comments:
```

"""contrib module containing StagingArea."""
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

from tensorflow.python.ops.data_flow_ops import StagingArea
```

PR #286, opened 2017-06-02
fix dependency: add python-opencv to requirements.txt

lispc:


lispc:
ok~ I will close the pull request~

PR #290, opened 2017-06-05
Add export for TensorFlow serving

PatWie:
Deploying models without the TensorPack dependency is sometime
useful or even necessary (TensorFlow serving). Example is also
provided.

[more details (why and how)](http://patwie.com/2017/06/05/intro_tensorpack_inference.html)


tensorpack/utils/export.py
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+
+import tensorflow as tf
+from tensorpack.utils import logger
+from tensorpack.tfutils import TowerContext
+from tensorpack.models import ModelDesc
+from tensorpack.tfutils import sessinit
+from tensorflow.python.saved_model import builder as saved_model_builder
+from tensorflow.python.saved_model import signature_constants, signature_def_utils, tag_constants, utils
```
ppwwyyxx: Since there are public APIs `tf.saved_model.builder` etc https://www.tensorflow.org/api_docs/python/tf/saved_model/builder/SavedModelBuilder, we can avoid these internal imports.

tensorpack/utils/export.py
```diff
@@ -0,0 +1,131 @@
...
+            input_names (list(str)): names of input tensors
+            output_names (list(str)): names of output tensors
+        """
+
+        assert isinstance(input_names, list)
+        assert isinstance(output_names, list)
+
+        logger.info('[export] prepare new model export')
+        super(ModelExport, self).__init__()
+        self.model = model()
```
ppwwyyxx: Why pass in a __class__ ? would it be more natural to just pass an instance?

ppwwyyxx:
Now `build(checkpoint)` can only be called once per instance. So this class is essentially one-time-use only.

I don't know if this use case really exists, but if you move the `checkpoint` argument into `export`, you can then do the following:
```python
e = Export(model, inputs, outputs)
e.build() # or do it in __init__
e.export(checkpoint1, export1)
# ...
e.export(checkpoint2, export2)
```

tensorpack/utils/export.py
```diff
@@ -0,0 +1,135 @@
...
+
+        logger.info('[export] prepare new model export')
+        super(ModelExport, self).__init__()
+        self.model = model
+        self.placehdrs = self.model.get_reused_placehdrs()
+        self.output_names = output_names
+        self.input_names = input_names
+
+    def export(self, checkpoint, export_path, version=1, tags=[tf.saved_model.tag_constants.SERVING],
+               signature_name='prediction_pipeline'):
```
ppwwyyxx: what is "version"?

ppwwyyxx:
One last thing, could you put the file in `tfutils/` instead?

PatWie:
You are wrong. The correct slogan is "One more thing (tm)".

ppwwyyxx:
What is "version"? Not used anywhere.

PatWie:
The old API had a versioning. I am pretty confident, they will add this also to the new one.

stephen-song:
@PatWie hi ，Can you re-deliver a link for **more details (why and how)** ?The previous link is no longer accessible.

PatWie:
The link is offline for a good reason: There are many changes in Tensor flow and I do not want to contribute another deprecated guide.

The file
https://github.com/tensorpack/tensorpack/blob/35beb43c4b30f703f885358fd9eb2776b47378ee/examples/basics/export-model.py

Should be pretty self-explanatory. If something changes, I will provide another PR to update this file whenever necessary. The serving part (Serverside) is up-to-date and covered as well under
https://github.com/PatWie/tensorflow-cmake/tree/master/serving

PR #294, opened 2017-06-06
Fix string decode issue for python3

PatWie:
A binary representation `"\n" * 5 + "hh".encode('utf-8')` reproduces
the error from travis-ci.


PatWie:
related log is: https://travis-ci.org/ppwwyyxx/tensorpack/jobs/239682860

Issue #266, opened 2017-05-11
allow changing location of train_log itself


eyaler:
log files are huge and i would like to change to location the train_log folder, say to a different drive.

PatWie:
You can use `logger.set_logger_dir(/mounts/trainlogs/myproject')` instead of `logger.auto_set_dir()`

eyaler:
thanks. would be nice to have an option in auto_set_dir to change 'train_log' but still use the base_name logic

ppwwyyxx:
`logger.set_logger_dir('/project/logs/' + os.path.basename(__file__))` would work fine.

PR #301, opened 2017-06-17
Add name to logger.autodir for different runs and simplify

PatWie:
I already had a previous pull-request about the same issue. This time summon the courage I try to convince you again about this feature. (see `mnist-embedding.py` for example). Sometimes, one uses the same training script (ResNet-18 vs. ResNet-50, mnist-embedding.py 'cosing' vs. mnist-embedding.py 'triplet'). But the runs all are tagged by the common scheme:

```
mynetwork0512-133348
mynetwork0515-094639
mynetwork0515-094924
mynetwork0515-095128
mynetwork0515-095232
mynetwork0515-095259
mynetwork0515-095339
mynetwork0515-095424
mynetwork0515-095619
mynetwork0515-095722
mynetwork0515-095751
mynetwork0515-100318
mynetwork0515-101107
mynetwork0515-101613
mynetwork0515-101841
mynetwork0515-102313
mynetwork0515-103231
```

while they are completely different. There are two other ways of solving this issue:
- use different files.py (ugly way duplicating code)
- always set the logger.dir manually by `logger.set_logger_dir`. But this does not give the interactive menu about `backup`, `new`, `delete`, ...

With `logger.auto_set_dir(name=some_value)` the list is much more understandable:

```
mynetwork:ideaA0512-133348
mynetwork:ideaA0515-094639
mynetwork:ideaA0515-094924
mynetwork:ideaB0515-095128
mynetwork:ideaB0515-095232
mynetwork:ideaB0515-095259
mynetwork:ideaC0515-095339
mynetwork:ideaC0515-095424
mynetwork:ideaC0515-095619
mynetwork:ideaC0515-095722
mynetwork:ideaC0515-095751
mynetwork:ideaD0515-100318
mynetwork:ideaD0515-101107
mynetwork:ideaB0515-101613
mynetwork:ideaD0515-101841
mynetwork:ideaD0515-102313
mynetwork:ideaA0515-103231
```

### DIFF
--- a/examples/SimilarityLearning/mnist-embeddings.py
+++ b/examples/SimilarityLearning/mnist-embeddings.py
@@ -210,7 +210,7 @@ def visualize(model_path, model):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('--gpu', help='comma separated list of GPU(s) to use.', required=True)
+    parser.add_argument('--gpu', help='comma separated list of GPU(s) to use.')
     parser.add_argument('--load', help='load model')
     parser.add_argument('-a', '--algorithm', help='used algorithm', type=str,
                         choices=["siamese", "cosine", "triplet", "softtriplet"])
--- a/examples/SimilarityLearning/mnist-embeddings.py
+++ b/examples/SimilarityLearning/mnist-embeddings.py
@@ -213,7 +213,7 @@ if __name__ == '__main__':
     parser.add_argument('--load', help='load model')
     parser.add_argument('-a', '--algorithm', help='used algorithm', type=str,
                         choices=["siamese", "cosine", "triplet", "softtriplet"])
-    parser.add_argument('--visualized', help='export embeddings into an image', action='store_true')
+    parser.add_argument('--visualize', help='export embeddings into an image', action='store_true')
     args = parser.parse_args()
 
     ALGO_CONFIGS = {"siamese": SiameseModel,
--- a/examples/SimilarityLearning/mnist-embeddings.py
+++ b/examples/SimilarityLearning/mnist-embeddings.py
@@ -9,7 +9,7 @@ import os
 from tensorpack import *
 import tensorpack.tfutils.symbolic_functions as symbf
 from tensorpack.tfutils.summary import add_moving_summary
-
+import argparse
 import tensorflow as tf
 import tensorflow.contrib.slim as slim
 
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
If `logger.set_logger_dir` sees an existing directory, it will still give an interactive menu

PatWie:
Do you really want us to write
```os.path.join('train_log', basename[:basename.rfind('.')])``` all the time instead of a parameter?

ppwwyyxx:
Of course it's harder to write with the dumb `set_logger_dir` interface. But "auto" is not so smart either. 
I can totally see that adding a "name" option would make things easier, but what if I have to store to a different directory (this is happening to me now because of disk space)? Another "dir" option? What if someone doesn't like "train_log" (#266 seems to suggest so)? Another option? Maybe the only thing people wouldn't hate is the basename part..

PatWie:
What about `TENSORPACK_LOGDIR`. You already have `TENSORPACK_DATASET` and `TENSORPACK_PIPES` or so.

ppwwyyxx:
That said, I'm OK with making a function for "getting the basename for the entry point" into utils. It's just 3 lines but not trivial lines..

To try new ideas I just rename the main script. I used to modify the scripts in-place until I realize it's hard to stay on track. I ran idea1-run1, idea2-run2... in parallel, and the next morning I forgot what exactly is idea1 and run1.

I've seen some scripts from openai, where every time something is run, a `git diff` is created to the log directory. That could be also a solution to my problem..

PatWie:
That `git diff` is only working, when one is regularly doing commits :-)

Ok, maybe the word `idea` is misleading.
Currently, I am directly facing a problem, where I can easily adjust a hyper-parameter about e.g. the depth of the network and I just find it convenient to have a function which adds a suffix to the path. Still, my current solution with pen and paper is not optimal.

ppwwyyxx:
An environment variable for log directory may not be a good idea, because I don't always want logs in the same place -- for small experiments I prefer them in the current directory. Env vars should be used for something that's always the same.

Imaging I don't have this function at all and someone sends a PR to add the same `auto_set_dir`, I certainly wouldn't like it because it's personal preference: current dir + "train_log" + basename. I think it's already a mistake to bake in my own preference -- clearly now it doesn't work for everyone (including you, #266, and me as well).

I don't know now what's the best thing to do about it. I guess one thing is to reduce the use of it in examples so others know the existence of `set_logger_dir`. Adding an option to partially fix the awkwardness might be OK, but the fix is very partial -- I still have the directory problem.

PatWie:
About the path issue: What about adding yet another argument "root" with Default "./train_log/"?

ppwwyyxx:
More options can be added of course.. but the point is that essentially people want to customize __every__ part in the path. Using the more general interface is the clean solution for users to understand what to do, rather than adding more and more options and explaining in the doc how the actual path is made from those options.

Anyway I think the name options is OK.

PR #322, opened 2017-07-01
Fix PTB example according to upstream changes

zsc:
Around March they included a stricter checking of variable scope, which leads to incompatibility with previous RNN implementations.
https://github.com/tensorflow/tensorflow/issues/8191

According to the example below, we need to set the reuse argument.
https://github.com/tensorflow/models/blob/master/tutorials/rnn/ptb/ptb_word_lm.py


ppwwyyxx:
Thanks!

PR #329, opened 2017-07-13
MultiGPU GAN Colocation Issue.

Skylion007:
Fixed bug with collocation of gradients in Tensorflow 1.1. It won't colocate the gradients because it complains /GPU: is not the same as /device:GPU . Changing this line fixes the spurious warning messages when training on multiple GPUs.


ppwwyyxx:
Opened an issue in tensorflow/tensorflow#11484.

ppwwyyxx:
A (better) work around is easy so I also pushed one.

PR #335, opened 2017-07-18
add _augment_coords implementation for many ImageAugmentor subclasses

haamoon:
I'll do the rest later. I have attached the test code (requires matplotlib and skimage). Let me know what you think. I'll add AugmentImageCoordinates shortly.

[test.zip](https://github.com/ppwwyyxx/tensorpack/files/1154688/test.zip)

### DIFF
--- a/tensorpack/dataflow/imgaug/crop.py
+++ b/tensorpack/dataflow/imgaug/crop.py
@@ -40,7 +40,10 @@ class RandomCrop(ImageAugmentor):
         return img[h0:h0 + self.crop_shape[0], w0:w0 + self.crop_shape[1]]
 
     def _augment_coords(self, coords, param):
-        raise NotImplementedError()
+        h0, w0 = param
+        coords[:, 0] = coords[:, 0] - w0
+        coords[:, 1] = coords[:, 1] - h0
+        return coords
 
 
 class CenterCrop(ImageAugmentor):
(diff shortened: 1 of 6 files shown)


tensorpack/dataflow/imgaug/crop.py
```diff
@@ -127,8 +137,10 @@ def _get_augment_params(self, img):
     def _augment(self, img, newbox):
         return newbox.roi(img)
 
-    def _augment_coords(self, coords, param):
-        raise NotImplementedError()
+    def _augment_coords(self, coords, newbox):
```
ppwwyyxx: Don't have to worry too much about these weird augmentors.. They are not quite well defined and not used by anyone, I think.

tensorpack/dataflow/imgaug/noname.py
```diff
@@ -70,6 +83,13 @@ def _augment(self, img, _):
             ret = ret[:, :, np.newaxis]
         return ret
 
+    def _augment_coords(self, coords, param):
+        h, w = param
+        #TODO Can self.interp value change these computation?
```
ppwwyyxx: I think interpolation doesn't affect the alignment in opencv.

tensorpack/dataflow/imgaug/crop.py
haamoon: I might need this in faster rcnn :)

tensorpack/dataflow/imgaug/noname.py
haamoon: Just deleted corresponding TODOs.

haamoon:
Just tested AugmentImageCoordinates. Use attached code to test all.
[Test.zip](https://github.com/ppwwyyxx/tensorpack/files/1154887/Test.zip)

tensorpack/dataflow/imgaug/crop.py
ppwwyyxx: Some sort of random bounding box perturbation is definitely helpful, but the problem is there isn't a standard way of doing this. If something isn't a standard, I think it's best to let users write their own version of it than to provide some arbitrary implementation. I might just mark this augmentor as deprecated.

haamoon:
Yes, that makes sense.

-------- Original message --------

tensorpack/dataflow/imgaug/crop.py
haamoon: Yes, that makes sense.

ppwwyyxx:
Thank you very much!

haamoon:
Absolutely. Let me know if you need help on other stuff. I would like to collaborate with you.

PR #387, opened 2017-08-19
add Affine geometry transformation

vfdev-5:
Hi, 
I would like to propose an affine image augmentor in `dataflow.imgaug.geometry`. [Here](https://gist.github.com/vfdev-5/f14490f0f8407f8f8b6b3035d159dd55) is an example of such transformations applied on one image. I will be glad to hear comments about this PR.

### DIFF
--- a/tensorpack/dataflow/imgaug/geometry.py
+++ b/tensorpack/dataflow/imgaug/geometry.py
@@ -8,7 +8,7 @@ import math
 import cv2
 import numpy as np
 
-__all__ = ['Shift', 'Rotation', 'RotationAndCropValid']
+__all__ = ['Shift', 'Rotation', 'RotationAndCropValid', 'Affine']
 
 
 class Shift(ImageAugmentor):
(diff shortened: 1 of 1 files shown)


tensorpack/dataflow/imgaug/geometry.py
```diff
@@ -150,3 +150,98 @@ def largest_rotated_rect(w, h, angle):
...
+
+        # Apply shear :
+        if self.shear > 0.0:
+            m00 = transform_matrix[0, 0]
+            m01 = transform_matrix[0, 1]
+            m10 = transform_matrix[1, 0]
+            m11 = transform_matrix[1, 1]
+            transform_matrix[0, 1] = m01 * cos_shear + m00 * sin_shear
+            transform_matrix[1, 1] = m11 * cos_shear + m10 * sin_shear
+            tx = center[0] * (1.0 - m00) - center[1] * transform_matrix[0, 1]
```
ppwwyyxx: What is this for?

vfdev-5: The lines 227 and 228:
```
transform_matrix[0, 1] = m01 * cos_shear + m00 * sin_shear
transform_matrix[1, 1] = m11 * cos_shear + m10 * sin_shear
```
inserts shear into rotational part : 
```
[[ scale * cos(a),    scale*sin(a+shear)]
[-scale*sin(a),       scale*cos(a+shear)]]
```
and yes, `tx` and `ty` keep the center unchanged with shear

Issue #386, opened 2017-08-19
freeze_graph gives truncated error after export


eyaler:
i am trying to run tensorflow freeze_graph after doing export.

```
        from tensorflow.python.tools import freeze_graph

        tf.reset_default_graph()
        e = ModelExport(Model(), input_names, output_names)
        e.export(model_file, export_dir)

        freeze_graph.freeze_graph(
            input_graph=os.path.join(export_dir, 'saved_model.pb'), input_saver="", input_binary=True,
            input_checkpoint=os.path.join(export_dir, 'variables', 'variables.data-00000-of-00001'),
            output_node_names="output_node", restore_op_name = "save/restore_all", filename_tensor_name = "save/Const:0",
            output_graph=os.path.join(export_dir, 'saved_model_frozen.pb'), clear_devices=True, initializer_nodes="")

```

i get 
```

  File "C:\Anaconda3\lib\site-packages\tensorflow\python\tools\freeze_graph.py", line 170, in freeze_graph
    input_graph_def = _parse_input_graph_proto(input_graph, input_binary)
  File "C:\Anaconda3\lib\site-packages\tensorflow\python\tools\freeze_graph.py", line 137, in _parse_input_graph_proto
    input_graph_def.ParseFromString(f.read())
  File "C:\Anaconda3\lib\site-packages\google\protobuf\message.py", line 185, in ParseFromString
    self.MergeFromString(serialized)
  File "C:\Anaconda3\lib\site-packages\google\protobuf\internal\python_message.py", line 1069, in MergeFromString
    if self._InternalParse(serialized, 0, length) != length:
  File "C:\Anaconda3\lib\site-packages\google\protobuf\internal\python_message.py", line 1105, in InternalParse
    pos = field_decoder(buffer, new_pos, end, self, field_dict)
  File "C:\Anaconda3\lib\site-packages\google\protobuf\internal\decoder.py", line 633, in DecodeField
    if value._InternalParse(buffer, pos, new_pos) != new_pos:
  File "C:\Anaconda3\lib\site-packages\google\protobuf\internal\python_message.py", line 1105, in InternalParse
    pos = field_decoder(buffer, new_pos, end, self, field_dict)
  File "C:\Anaconda3\lib\site-packages\google\protobuf\internal\decoder.py", line 612, in DecodeRepeatedField
    if value.add()._InternalParse(buffer, pos, new_pos) != new_pos:
[...]

PR #400, opened 2017-08-31
Add centerloss and refactor symbf

PatWie:
symbolic functions.py should be slim as for most projects there is no
need for the entire collection of embedding losses.


ppwwyyxx:
Don't know exactly why, but running `gm convert results.jpg new.jpg` makes the image half the size without visible quality change (https://github.com/ppwwyyxx/tensorpack/pull/400/commits/12648f22c69393447cb2f1aab4746ce8a76b65fe)

PatWie:
I used opencv to combine the old and new one. Too lazy to run all algorithms again.

Skylion007:
Ah man, I was using those symbolic functions. They are actually incredibly useful for any type of Siamese networks, I would recommend commonly used metrics/model types should be better stored in Tensorpack. For instance, does GAN.py with it's useful and specific trainers really belong in examples?

ppwwyyxx:
Apologies for the break. I should've marked them as deprecated first.
They're removed for the same reasons why GAN is not included or [freeze graph](https://github.com/ppwwyyxx/tensorpack/issues/386#issuecomment-324694820) is not included. Being useful is never the whole reason to add something.

PatWie:
I started a `loss.py` file. But I do not want to maintain it and seems to be not in line with the decision to use `tf.layers`. But it might be worth at thinking about such a `loss.py` file as TensorFlow only supports streaming-metrics (not loss functions).

ppwwyyxx:
TF has a few in `tf.losses.xxx` and a few in `tf.nn.xxx_loss`. But yes they are very limited.
However I do need to remove `huber_loss` and use `tf.losses.huber_loss`.

Having some symbolic function collections is never a bad idea but tensorpack isn't the right place for it. I'd like to see a project called "a bunch of useful TF symbolic functions" and then not only tensorpack but also anyone who uses tensorflow could benefit from it.

Skylion007:
Alternatively, someone could create that project and have it be a dependency of Tensorpack. While Tensorpack may not be the proper place for many of these loss functions, it has by far some of the cleanest and best implementations of them. I'd be happy to try to start that project if there is interest. @PatWie care to share a gist of the loss.py file?

Regardless, I wouldn't mind seeing loss.py as part of say the example section.

PR #481, opened 2017-11-09
cifar-resnet-mixup implement

dongzhuoyao:
origin cifar result is 5.73 in my experiment, this mixup result is 5.23.  detail result summay will be later supplied.


scripts/dump-model-params.py
```diff
@@ -1,59 +0,0 @@
-#!/usr/bin/env python
```
ppwwyyxx: Could you get the scripts back?

tensorpack/dataflow/common.py
```diff
@@ -149,6 +149,60 @@ def _aggregate_batch(data_holder, use_list=False):
         return result
 
 
+class BatchDataByMixup(BatchData):
```
ppwwyyxx: Could you put this in the example script? And it would seem easier to write it as a `MapData` on top of a batch.

dongzhuoyao: yes, i can do that, but in that way I still need hack the MapData method in the "common.py", because mixup needs two data, however, the MapData only returns one. any good suggestion?

ppwwyyxx: What you've written is literally `MapData(BatchData(ds, batch_size), mixup)`.

examples/ResNet/cifar10-resnet-mixup.py
```diff
@@ -0,0 +1,205 @@
...
+        x1 = images[:real_batch_size]
+        x2 = images[real_batch_size:]
+        x = x1 * x_weight + x2 * (1 - x_weight)
+        y1 = one_hot_labels[:real_batch_size]
+        y2 = one_hot_labels[real_batch_size:]
+        y = y1 * y_weight + y2 * (1 - y_weight)
+        return [x, y]
+        return ds
+
+    ds = BatchData(ds, 2*BATCH_SIZE, remainder=not isTrain)
```
ppwwyyxx: mixup data should be inside "if isTrain".

ppwwyyxx:
In the paper, "CIFAR-10 PreAct ResNet-18" error goes from 5.6 to 3.9

dongzhuoyao:
I will check this "5.6 vs 3.9" problem.

dongzhuoyao:
if I set alpha=1, I can obtain a 4.74 result. could you merge the code first?  

I will setup a README file to clarify the difference between the paper. hope me or someone else interested can fix it later.

ppwwyyxx:
Sorry but we hope to only include examples with [reproduced performance](https://github.com/ppwwyyxx/tensorpack/tree/master/examples#speech--nlp). It's hard and that's why it's important. Otherwise it becomes a big collection of code that appears useful but frustrates the users, like most of the deep learning code I've seen on github. Pretty much everyone knows how to write a new model, but the hard part is in achieving the performance.
In anycase a README is a must. Without a result that people can reproduce, open source deep learning code is of no meaning.

ppwwyyxx:
Briefly talked to the author. Their "CIFAR-10 PreAct ResNet-18" is the architecture [here](https://github.com/kuangliu/pytorch-cifar/blob/master/models/preact_resnet.py), not exactly what we have in the examples.

dongzhuoyao:
the new preact18 architecture achieves 4.08, still exists a 1.08 difference.   =。=

ppwwyyxx:
Cool. What's the number without mixup?

dongzhuoyao:
haven't done it, will do it later.

dongzhuoyao:
preact18-origin: 5.08
preact18-mixup:3.7

is this result ok?

ppwwyyxx:
Good result! Is this still following the kuangliu/pytorch-cifar setting?

dongzhuoyao:
yes, the code now  is a little messy,  I will clean it up

ppwwyyxx:
Sure. Please also change the docstrings at top of the file, to include notes about the settings and the performance.

dongzhuoyao:
ok

dongzhuoyao:
updated

ppwwyyxx:
Could you merge two scripts into one with a flag? For example purposes it doesn't make sense to have many files with almost identical contents.

dongzhuoyao:
done

ppwwyyxx:
Each resnet block in the reference code has conv{1,2,3} and another conv in shortcut. https://github.com/kuangliu/pytorch-cifar/blob/886af4caa4ee224c3fd4b42e997e43b1b4f234b9/models/preact_resnet.py#L57-L64 
It seems to be missing here?

dongzhuoyao:
sorry for my carelessness, even though got a 3.7 result.  is this code ok? I will retrain it later.

ppwwyyxx:
Thanks! The code is OK. If you don't have time I can train it as well.

dongzhuoyao:
already training. :smile:

dongzhuoyao:
updated

Issue #489, opened 2017-11-13
allow image augmentations to throw exception


eyaler:
image augmentations e.g. AugmentImageComponent swallow all non-keyboard exceptions. while they do output a printout, this sometimes goes unnoticed due to later printouts of my program thus effectively failing silently. if this stays the default behaviour i would like an option to override it

ppwwyyxx:
The original idea was to allow some errors to happen occasionally as long as the training is still going, because failing the whole training due to some dirty data can be frustrating.
But I think this is not a good default, due to reasons you mentioned. You can add an option and change the default behavior. Do you want to work on it?

ppwwyyxx:
The current exception handling code piece is repeated for 3 times -- which means it needs to be extracted out to a common function, with an option that disables exception handling.

eyaler:
taking

PR #490, opened 2017-11-13
enable image exceptions and make this default behavior

eyaler:
https://github.com/ppwwyyxx/tensorpack/issues/489

### DIFF
--- a/tensorpack/dataflow/image.py
+++ b/tensorpack/dataflow/image.py
@@ -18,6 +18,7 @@ def _valid_coords(coords):
     assert coords.shape[1] == 2, coords.shape
     assert np.issubdtype(coords.dtype, np.float), coords.dtype
 
+
 class ExceptionHandler:
     def __init__(self, catch_exceptions=False):
         self._nr_error = 0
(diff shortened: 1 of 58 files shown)


tensorpack/dataflow/image.py
```diff
@@ -57,7 +68,7 @@ class AugmentImageComponent(MapDataComponent):
     """
     Apply image augmentors on 1 image component.
     """
-    def __init__(self, ds, augmentors, index=0, copy=True):
+    def __init__(self, ds, augmentors, index=0, copy=True, allow_exceptions=False):
```
ppwwyyxx: The name `allow_exceptions` confused me at the first look. The current code is saying exception will be raised when exception is "not allowed", which is weird.
Maybe use "hide_exceptions" or "catch_exceptions"?

tensorpack/dataflow/image.py
```diff
@@ -17,6 +17,17 @@ def _valid_coords(coords):
     assert coords.shape[1] == 2, coords.shape
     assert np.issubdtype(coords.dtype, np.float), coords.dtype
 
+class ExceptionHandler:
+    def __init__(self, allow_exceptions=False):
+        self._nr_error = 0
+        self.allow_exceptions = allow_exceptions
+
+    def catch(self):
```
ppwwyyxx: You can make it a contextmanager, so the `try ... catch` logic can be shared as well. 
An example: https://github.com/ppwwyyxx/tensorpack/blob/2fa49895399ec33298f2b774929aada5fbed3e89/tensorpack/dataflow/prefetch.py#L51-L65.

tensorpack/dataflow/image.py
eyaler: -> catch_exceptions

PR #498, opened 2017-11-18
"Let there be color"

PatWie:
As I promised in our first "TensorPack-meeting", I will start contributing again after the CVPR deadline.

I found the Siggraph paper:
_Let there be Color!: Joint End-to-end Learning of Global and Local Image Priors for Automatic Image Colorization with Simultaneous Classification_
 <http://hi.cs.waseda.ac.jp/~iizuka/projects/colorization/en/>

relatively straight-forward to implement. Please note, I haven't trained it until the end (still ongoing).

Not sure about the results. In some cases, I prefer the predicted images over the ground-truth.

## some failure cases
failure0
failure1

## working cases
prediction
prediction2
prediction3
prediction4

Please consider this as a first draft. I will clean up the implementation when training is done.


PatWie:
comparison

colorize
color2

I think, that this model is equally good at the Place Dataset, but does not generalize as good as the authors model. Seems to be reasonable, it is their teaser image which usually shows the best results. Further, there is some resizing going on in their inference code and they filtered out some images from the Place-Dataset, without describing details. And of course, I did not train the model for 3 weeks.

Any suggestions?

PatWie:
changed status from WIP -> "I do not want to touch it again"

PatWie:
I cleaned the README and the pre-trained model is [here](files.patwie.com/models/let-there-be-color.npy).

PR #504, opened 2017-11-22
add VGG19

PatWie:
I would upload the converted (and tested) VGG-19 model as well.


PatWie:
I did not find a Github gist for the vgg19-deploy.proto of the caffe-model (only train_val).

ppwwyyxx:
Having both load-vgg16 and load-vgg19 still feels weird. Maybe later will move things (with load-alexnet, or maybe load-cpm as well) to a separate directory.

PatWie:
As this is commonly used, I am not sure if we should merge them into a single file hide the model behind flags and if-branches. But a good alternative would be having `VGG16Model` and `VGG19Model`.

PatWie:
Ok too late :grin:

ppwwyyxx:
Ideally similar code should not co-exist, like I'm saying in another PR: https://github.com/ppwwyyxx/tensorpack/pull/481#issuecomment-345672569
And having two models (or just two tower_func) sounds good.
But you're right, for something like VGG, people usually just want something simple to copy and paste..

PR #510, opened 2017-11-23
allow prefix for dictrestore

PatWie:
When using a pretrained model like VGG under a different variable_scope, currently only `SaverRestore` supports a prefix. This commit adds this functions for DictRestore as well.

### DIFF
--- a/tensorpack/tfutils/sessinit.py
+++ b/tensorpack/tfutils/sessinit.py
@@ -186,7 +186,7 @@ class DictRestore(SessionInit):
     Restore variables from a dictionary.
     """
 
-    def __init__(self, variable_dict):
+    def __init__(self, variable_dict, prefix=None):
         """
         Args:
             variable_dict (dict): a dict of {name: value}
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
For `DictRestore` this is not needed. Users can just use
`dict = {prefix + name: value for name, value in six.iteritems(dict)}` to do the same thing.

PatWie:
Users dont care about the internal format of npy files. They want to restore weights from npy files like restoring from checkpoint. It is a matter of consistency.

ppwwyyxx:
Users should care about the format if they use DictRestore. That's why it takes a dict but not a file name.

ppwwyyxx:
In fact using the dict and npy format is specifically for users to load and modify the model manually. Being easy to load and modify is the only benefit it has compared to tf format.

PatWie:
I see your point. Just wondering, if `class *Restore` shouldn't share the same interface.

dongzhuoyao:
it may be better discuss the function first,  then start coding.

PatWie:
I just code them, when **I need** them without asking anybody.
Honestly, I don't care if @ppwwyyxx wants them to be included a public version of tensorpack or not. If not, I close the pull-request, like I did in the past. That's the nice thing about GIT and forks.

I understand the point, that this library wants to be light-weight. But I this case I am still convinced, it might be good to keep the interfaces at least consistent. So I use it in my local version. 

I respect @ppwwyyxx decision. So there should be no discussion.

ppwwyyxx:
About forks, you can also just write a collection of functions for your convenience, e.g. a wrapper around DictRestore or logger.set_logger.dir. This might be easier than maintaining a personal fork.

About light-weight. Yes I hope it's lightweight -- though there is a lot of complexity already. So in general I don't like one-line features. 

About consistency: if there are 5 *Restore and one is different from others, I'll call it a consistency issue. These two in my mind were just two things with different functionalities and therefore different APIs. I don't even know if there will be a third *Restore, because I can't imagine what else one can need.

PR #514, opened 2017-11-25
Fix se_resnet_bottleneck to support NHWC data format

sirotenko:


examples/ResNet/resnet_model.py
```diff
@@ -97,6 +97,11 @@ def se_resnet_bottleneck(l, ch_out, stride):
     squeeze = FullyConnected('fc1', squeeze, ch_out // 4, nl=tf.nn.relu)
     squeeze = FullyConnected('fc2', squeeze, ch_out * 4, nl=tf.nn.sigmoid)
     l = l * tf.reshape(squeeze, [-1, ch_out * 4, 1, 1])
```
ppwwyyxx: remove the original reshape

PR #522, opened 2017-11-29
Add Real-Time Single Image Superresolution (WIP)

PatWie:
Real-Time Single Image and Video Super-Resolution Using an Efficient Sub-Pixel Convolutional Neural Network
Wenzhe Shi, et al. https://arxiv.org/abs/1609.05158


ppwwyyxx:
When I played with SR before I found opencv bicubic has very bad PSNR compared to matlab bicubic, which makes some results very hard to reproduce. Maybe you want to check that.

PatWie:
Most of these papers kind of cheat by blurring the highres images before bi-cubic downscaling the image for a low-res input. If so, the gap between bi-cubic and network visually _seems_ to be larger than here. This Gaussian blur usually helps the network as the information from the highres is partly available in the lowres.
 Further, the public benchmarks I found, use gray-scale images. 

I observed the opposite (which agrees to https://stackoverflow.com/a/22093004). But I guess, this again depends on the specific input :-) I was a little disappointed seeing these piblic benchmarks only contain very few images....

ppwwyyxx:
```

opencv:
import numpy as np
import cv2
import sys
im = cv2.imread(sys.argv[1])
im = im.astype('float32')
imy = cv2.cvtColor(im, cv2.COLOR_BGR2YCR_CB)
print imy.shape
imy = imy[:,:,0]
down = cv2.resize(imy, (0,0), 0, 0.5, 0.5, cv2.INTER_CUBIC)
print down.shape
up = cv2.resize(down, (0,0), 0, 2, 2, cv2.INTER_CUBIC)
v = np.square(up - imy).mean()
print v
print 20 * np.log10(255 / np.sqrt(v))


matlab:
function bicubic
im = imread('barbara.bmp');
im = rgb2ycbcr(im);
imy = im(:,:,1);
up_scale=2;
down = imresize(imy, 0.5, 'bicubic');
up = imresize(down, 2, 'bicubic');
diff = up - imy;
l2 = mean(mean(diff.*diff))
20 * log10(255/sqrt(l2))
end
```
This was the test code I was using (two years ago when I tried to reproduce SR-CNN). It only compares the y-channel. I remember on Set14 opencv is much worse than matlab.

PatWie:
I do not have the time to test the entire dataset (I took enough to install Matlab R2016a), but for the first example I took:
[comparison.tar.gz](https://github.com/ppwwyyxx/tensorpack/files/1517114/comparison.tar.gz)
python is slightly better:

```
python 25.3957339631 db
matlab 25.0888434195 db
```

But I agree
compare

the Matlab results looks much better (although it has a lower PSNR)

PatWie:
Interestingly enough `from Image import PIL` gives significantly better performance

```
python 25.3957339631 db
matlab 25.0888434195 db
pil    26.8381044874 db
```

PR #527, opened 2017-11-30
add enhancenet

PatWie:
After tweaking the loss function parameters, I eventually got some comparable results. 

But the training was **highly highly highly** unstable. I slightly altered the learning rate and some parameters, which appears to be more stable. Further, the paper is a little bit vague about batch-size, normalization steps, and the generator/discriminator balancing:

The article states that training was done by balancing the generator/discriminator loss functions (skipping discriminator updates, if too good) which is listed [here at 11](https://github.com/soumith/ganhacks) as a don't do. I just tweak the hyperparameters.

Click on the image, for better comparison.

enhancenet-demo

The network training is still running.  I can share the weights for this implementation.


PatWie:
I might be easy to further convert the release model by the authors into a python dictionary, which can be restored by `DictRestore`. I haven't tried this yet and just used the author's code.

PatWie:
Results are available and comparable with the authors implementations (there is no clear winner)
[results.tar.gz](https://github.com/ppwwyyxx/tensorpack/files/1522218/results.tar.gz)

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,334 @@
...
+
+        if not self.inference:
+            with tf.variable_scope('discrim'):
+                real_score = discriminator(real_hr)
+                fake_score = discriminator(fake_hr)
+
+            self.build_losses(real_score, fake_score)
+
+            if ctx.is_training:
+                additional_losses = additional_losses(fake_hr, real_hr)
```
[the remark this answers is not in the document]
PatWie: works without this multiplication

PatWie:
I think this pull request is ready to be reviewed. My trained model is here:
http://files.patwie.com/models/enet-pat.npy
Feel free to add suggestions or edit when required.

Please consider, I am not that patient to train it more epochs and cherry pick the best model.

ppwwyyxx:
Do you think you need `align_corners=True` for `tf.image.resize_*`? 
To be honest I don't like `tf.image.resize_` because they never align pixel to pixel regardless of `align_corners`. But maybe True makes more sense here?

You can check this out:
```python
import numpy as np
from skimage.transform import rescale

np.set_printoptions(linewidth=10000)

arr = np.array(
    [[1,2,3,4],
    [5,6,7,8],
    [9,10,11,12]], dtype='float32')

input = tf.constant(arr)
input4D = tf.reshape(input, [1, 3, 4, 1])
resize = tf.image.resize_bilinear(input4D, [6, 8], align_corners=True)[0,:,:,0]
sess = tf.Session()
r1 = sess.run(resize)
r2 = rescale(arr/100.0, 2, mode='edge') * 100

print(r1)
print(r2)
```

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,334 @@
...
+            _, h, w, _ = x.get_shape().as_list()
+            x = tf.image.resize_nearest_neighbor(x, [factor * h, factor * w])
+            return x
+
+        def add_VGGMeans(x):
+            with tf.name_scope('add_vgg_means'):
+                r, g, b = tf.unstack(x, axis=3)
+                r += 123.68 / 255.
+                g += 116.779 / 255.
+                b += 103.939 / 255.
```
ppwwyyxx: use the constants

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,334 @@
...
+    assert isinstance(v, tf.Tensor)
+    v.get_shape().assert_has_rank(4)
+    dim = v.get_shape().as_list()
+    v = normalize(v)
+    v = tf.reshape(v, [dim[0], dim[1] * dim[2], dim[3]])
+    return tf.matmul(v, v, transpose_a=True)
+
+
+def mse(x, y, name=None):
+    return tf.reduce_mean(tf.squared_difference(x, y), name=name)
```
ppwwyyxx: tf.losses.mean_squared_error(reduction=Reduction.MEAN)

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,334 @@
...
+            # texture loss
+            with tf.name_scope('texture_loss'):
+                def texture_loss(x, p=16):
+                    x = normalize(x)
+                    _, h, w, _ = x.get_shape().as_list()
+                    rsl = []
+                    logger.info('create texture loss for layer {} with shape {}'.format(
+                        x.name, x.get_shape()))
+                    for i in range(0, h - p, p - 1):
+                        for j in range(0, w - p, p - 1):
```
ppwwyyxx: tf.space_to_batch_nd + reshape ?

examples/Superresolution/enet-pat.py
PatWie: fixed

examples/Superresolution/enet-pat.py
PatWie: I feel more comfortable having the loop, which is IMHO much easier to parse than hiding this behind some functions. But I am fine with `tf.space_to_batch_nd` if this is equivalent.

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,334 @@
...
+                tf.summary.image('input,fake,real', viz,
+                                 max_outputs=max(30, BATCH_SIZE))
+
+            self.collect_variables()
+
+    def _get_optimizer(self):
+        lr = tf.get_variable(
+            'learning_rate', initializer=1e-4, trainable=False)
+        opt = tf.train.AdamOptimizer(lr)
+        gradprocs = [gradproc.ScaleGradient([('VGG19/*', 0.)]), gradproc.ScaleGradient([('discrim/*', 0.3)])]
```
ppwwyyxx: use tf.stop_gradient for VGG?
Even if lr=0 gradients are still computed and wasted

PatWie:
TensorFlow does very strange things when resizing images  with/without `align_corners`. For the sake of consistency I would like to keep the current version. 
Feel free to retrain the model :wink:

examples/Superresolution/enet-pat.py
PatWie: any other solution?

examples/Superresolution/enet-pat.py
ppwwyyxx: And why `p-1` instead of `p`?

PatWie: To get all patches? When using `p` it skips the last patch in each row/column.

ppwwyyxx: Perhaps. Just feel looping over pixels isn't very tensorflowish.
```python
x = tf.space_to_batch_nd(x, [p, p], [[0,0],[0,0]])
x = tf.reshape(x, [p, p, -1, h, w, c])
x = tf.transpose(x, [2, 0, 1, 3, 4, 5])
patches_a, patches_b = tf.split(x, 2)   # each is b,p,p,h/p,w/p,c
```

examples/Superresolution/enet-pat.py
ppwwyyxx: `with varreplace.freeze_variables():`

PatWie: Does this affects the gradients wrt. to `fake_hr`?

examples/Superresolution/enet-pat.py
ppwwyyxx: Shouldn't it be `range(0, h-p+1, p)`?

ppwwyyxx: Or equivalently range(0, h, p), when h%p==0

PatWie: Spiced with a pinch of pytorch style. Honestly, I was not aware of `tf.space_to_batch_nd`.

ppwwyyxx: But using space_to_batch_nd fixes this anyway, so..

examples/Superresolution/enet-pat.py
ppwwyyxx: No it'll only stops the gradients on variables (inside the with block), not on any layer output.

examples/Superresolution/enet-pat.py
PatWie: Well spotted! I don't think this has a huge impact. 
Even when 128/16=8, I would prefer `space_to_batch_nd`.

examples/Superresolution/enet-pat.py
PatWie: That's awesome!

examples/Superresolution/enet-pat.py
ppwwyyxx: fix:
```python
x = tf.space_to_batch_nd(x, [p, p], [[0,0],[0,0]])
x = tf.reshape(x, [p, p, -1, h/p, w/p, c])
x = tf.transpose(x, [2, 3, 4, 0, 1, 5])
patches_a, patches_b = tf.split(x, 2)   # each is b,h/p,w/p,p,p,c
```

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,330 @@
...
+    return tf.matmul(v, v, transpose_a=True)
+
+
+class Model(GANModelDesc):
+
+    def __init__(self, height=SHAPE_LR, width=SHAPE_LR, inference=False):
+        super(Model, self).__init__()
+        self.height = height
+        self.width = width
+        self.inference = inference
```
ppwwyyxx: why this? seems like `inference=ctx.is_training` is always true

examples/Superresolution/enet-pat.py
```diff
@@ -0,0 +1,330 @@
...
+
+    pred = predict_func(lr[None, ...])
+    p = np.clip(pred[0][0, ...] * 255, 0, 255)
+
+    cv2.imwrite(output_path + "predition.png", p)
+    cv2.imwrite(output_path + "baseline.png", baseline)
+
+
+def get_data(lmdb):
+    ds = LMDBDataPoint(lmdb, shuffle=True)
```
ppwwyyxx: Is this faster than putting `ImageDataFromZIPFile` + preprocess here?
I guess maybe just the `zipfile` module is really slow. You are doing random read with LMDB anyway.

ppwwyyxx:
Is the visualization produced by Saccade?

examples/Superresolution/enet-pat.py
PatWie: not during "apply". There is a difference subtle difference between a callback inference runner and the `apply` function. The inference runner might need the loss statistics, the `appply` function not.

examples/Superresolution/enet-pat.py
PatWie: At some point I did benchmarks.

PatWie:
The visualization is pure LaTeX + Tikz

examples/Superresolution/enet-pat.py
ppwwyyxx: OfflinePredictor always creates model with ctx.is_training=False

PatWie: InferenceRunnerCallback?

ppwwyyxx: I understand the need, but you're not using inferencerunner

examples/Superresolution/data_sampler.py
```diff
@@ -0,0 +1,123 @@
...
+import numpy as np
+import zipfile
+import random
+from tensorpack import RNGDataFlow, MapDataComponent, dftools
+
+"""
+DBDIR=/graphics/projects/scratch/datasets/mscoco/
+python data_sampler.py --lmdb ${DBDIR}val2017.lmdb --input ${DBDIR}val2017.zip --create
+python data_sampler.py --lmdb ${DBDIR}train2017.lmdb --input ${DBDIR}train2017.zip --create
+"""
```
ppwwyyxx: Instructions are in readme already. better to not duplicate them

examples/Superresolution/data_sampler.py
```diff
@@ -0,0 +1,123 @@
...
+
+# Testcode for encode/decode.
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--create', action='store_true', help='create lmdb')
+    parser.add_argument('--debug', action='store_true', help='debug images')
+    parser.add_argument('--input', type=str, help='path to lmdb', required=True)
+    parser.add_argument('--lmdb', type=str, help='path to lmdb', required=True)
+    parser.add_argument('--size', type=int, help='number of images', default=1000)
+    parser.add_argument('--proc', type=int, help='number of processes', default=1)
```
ppwwyyxx: unused?

ppwwyyxx:
At last, rename to `SuperResolution`

examples/Superresolution/data_sampler.py
PatWie: Oops forgot to remove them

PatWie:
Should I do these changes, or did you started it already?

examples/Superresolution/data_sampler.py
PatWie: fixed

examples/Superresolution/data_sampler.py
PatWie: fixed

examples/Superresolution/enet-pat.py
PatWie: fixed (as I do not use the InferenceRunner here). Maybe use `ctx.phase = ['train', 'val', 'inference']` at some point

examples/Superresolution/enet-pat.py
PatWie: Haven't changed it, but then
```python
x = tf.space_to_batch_nd(x, [p, p], [[0,0],[0,0]])
x = tf.reshape(x, [p, p, -1, h/p, w/p, c])
x = tf.transpose(x, [2, 3, 4, 0, 1, 5])
patches_a, patches_b = tf.split(x, 2)   # each is b,h/p,w/p,p,p,c

patches_a = tf.reshape(patches_a, [-1, p, p, c])
patches_b = tf.reshape(patches_b, [-1, p, p, c])
```
should work

examples/Superresolution/enet-pat.py
PatWie: fixed

examples/Superresolution/enet-pat.py
PatWie: fixed by  `range(0, h, p)`, `range(0, w, p)`

examples/Superresolution/enet-pat.py
ppwwyyxx: Well yeah, I also felt having these two phases isn't perfect, though users can always do their own stuff like you did.

PatWie: I had this issue not before. As long as GANs are not stable (requiring the InferenceRunner), we do not need these different phases :-)

examples/SuperResolution/enet-pat.py
```diff
@@ -0,0 +1,310 @@
...
+    v.get_shape().assert_has_rank(4)
+    dim = v.get_shape().as_list()
+    return v / (dim[1] * dim[2] * dim[3])
+
+
+def gram_matrix(v):
+    assert isinstance(v, tf.Tensor)
+    v.get_shape().assert_has_rank(4)
+    dim = v.get_shape().as_list()
+    v = normalize(v)
```
ppwwyyxx: The texture loss seems too small to be useful (there is already a very small loss weight for them). Does it really average over these dimensions here? The paper only mentions L2 norm.

ppwwyyxx: "For the perceptual loss LP and the texture loss LT , we
normalized feature activations to have a __mean__ of one [4]."

ppwwyyxx:
What is "enet-pat"....? will you rename it to "enhancenet"?

ppwwyyxx:
And how long did you train these models?

PatWie:
"enet-pat" (patch-, adversarial, & texture-loss) is from the paper, see Figure 3 in the paper. I trained them with two gpus (hence batchsize 12) until global step 806817 (I use the checkpoint model-806817).

examples/SuperResolution/enet-pat.py
```diff
@@ -0,0 +1,310 @@
...
+def get_data(lmdb):
+    ds = LMDBDataPoint(lmdb, shuffle=True)
+    ds = ImageDecode(ds, index=0)
+    augmentors = [imgaug.RandomCrop(128),
+                  imgaug.Flip(horiz=True)]
+    ds = AugmentImageComponent(ds, augmentors, index=0, copy=True)
+    ds = MapData(ds, lambda x: [np.stack([x[0][:, :, 0] - VGG_MEAN[0],
+                                          x[0][:, :, 1] - VGG_MEAN[1],
+                                          x[0][:, :, 2] - VGG_MEAN[2]], axis=-1)])
+    ds = MapData(ds, lambda x: [cv2.resize(x[0], (32, 32), interpolation=cv2.INTER_AREA), x[0]])
```
ppwwyyxx: why "area"?

ppwwyyxx:
Why would you have `checkpoint model-806817`? Are we using the same dataset?:
```
100%|#######################|19714/19714[59:02<00:00, 5.56it/s]          
[1206 04:49:54 @base.py:255] Epoch 40 (global_step 788560) finished, time:3542.78 sec.                                                             
[1206 04:49:55 @saver.py:82] Model saved to train_log/enet-pat/model-788560.                                                                       
[1206 04:49:55 @monitor.py:363] GAN_loss/discrim/accuracy: 0.99414       
[1206 04:49:55 @monitor.py:363] GAN_loss/discrim/loss: 0.019299          
[1206 04:49:55 @monitor.py:363] GAN_loss/gen/accuracy: 0.0026996         
[1206 04:49:55 @monitor.py:363] GAN_loss/gen/loss: 8.9464
[1206 04:49:55 @monitor.py:363] QueueInput/queue_size: 49.585
[1206 04:49:55 @monitor.py:363] add: 19.204
[1206 04:49:55 @monitor.py:363] additional_losses/loss_LA: 8.9464
[1206 04:49:55 @monitor.py:363] additional_losses/loss_LP1: 1.3098
[1206 04:49:55 @monitor.py:363] additional_losses/loss_LP2: 0.0015531
[1206 04:49:55 @monitor.py:363] additional_losses/loss_LT1: 2.3112e-37
[1206 04:49:55 @monitor.py:363] additional_losses/loss_LT2: 2.2718e-37
[1206 04:49:55 @monitor.py:363] additional_losses/loss_LT3: 2.3154e-37
[1206 04:49:55 @base.py:245] Start Epoch 41 ...
100%|#######################|19714/19714[59:03<00:00, 5.56it/s]
[1206 05:48:58 @base.py:255] Epoch 41 (global_step 808274) finished, time:3543.57 sec.
```

examples/SuperResolution/enet-pat.py
ppwwyyxx: @PatWie

ppwwyyxx:
vgg19 should be applied to images of range [0,255]-mean. But just found that here it is on images of [0,1]-mean/255. (check the [initial commit](https://github.com/ppwwyyxx/tensorpack/pull/527/commits/b10effc0bff7882d8ec18af231f7974247e6d70b) )So the entire additional_losses would probably do nothing good.

PR #539, opened 2017-12-06
additional results and links

PatWie:
use the model from models.tensorpack.com for much better results and add ref in markdown files


PatWie:
The results are clearly better, though the training is not that stable.

PR #561, opened 2017-12-20
More verbose warning

PatWie:
For some reason, I  encounter this problem again and again. I see this assert 2 lines below, but the error msg can be a little bit more verbose.

### DIFF
--- a/tensorpack/models/registry.py
+++ b/tensorpack/models/registry.py
@@ -82,7 +82,8 @@ def layer_register(
             if use_scope:
                 name, inputs = args[0], args[1]
                 args = args[1:]  # actual positional args used to call func
-                assert isinstance(name, six.string_types), name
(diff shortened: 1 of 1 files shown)


tensorpack/models/registry.py
```diff
@@ -80,6 +80,9 @@ def wrapper(func):
         def wrapped_func(*args, **kwargs):
             assert args[0] is not None, args
             if use_scope:
+                if not isinstance(args[0], six.string_types):
+                    logger.warn("First argument for {} should be a string. "
+                                "Did you forget to specify the name of the layer?".format(func.__name__))
                 name, inputs = args[0], args[1]
                 args = args[1:]  # actual positional args used to call func
                 assert isinstance(name, six.string_types), name
```
ppwwyyxx: You can put the message here on the assertion.

PatWie:
That's better. You are right. I just was looking for red warnings.

PR #571, opened 2017-12-26
ResNet-mixup example: align implementation with one, referenced by original paper.

yselivonchyk:
ResNet-18 with preactivation as by https://github.com/kuangliu/pytorch-cifar is using ResNet with preactivation block with 2 consecutive convolution layers in the block. Existing implementation was using 3 convolutional blocks.

Exponentially decaying WEIGHT_DECAY replaced with fixed reference value.

Architecture aligned with main repository approach: defined functions for bottleneck and regular PreActResNet blocks.

Support for multiple depths added.

### DIFF
--- a/examples/ResNet/cifar10-preact18-mixup.py
+++ b/examples/ResNet/cifar10-preact18-mixup.py
@@ -59,8 +59,8 @@ RESNET_CONFIG = {
 
 FILTER_SIZES = [64, 128, 256, 512]
 
+
 def preactivation_block(input, num_filters, stride=1):
-    net = BNReLU(input)
     num_filters_in = net.get_shape().as_list()[1]
     # identity
     shortcut = net
@@ -68,7 +68,8 @@ def preactivation_block(input, num_filters, stride=1):
         shortcut = Conv2D('shortcut', net, num_filters, kernel_shape=1, stride=stride, use_bias=False,
                           nl=tf.identity)
     # residual
-    residual = Conv2D('conv1', net, num_filters, kernel_shape=3, stride=stride, use_bias=False, nl=BNReLU)
+    residual = BNReLU(input)
+    residual = Conv2D('conv1', residual, num_filters, kernel_shape=3, stride=stride, use_bias=False, nl=BNReLU)
     residual = Conv2D('conv2', residual, num_filters, kernel_shape=3, stride=1, use_bias=False, nl=tf.identity)
     return shortcut + residual
 
@@ -76,15 +77,15 @@ def preactivation_block(input, num_filters, stride=1):
 def bottleneck_block(input, num_filters, stride=1):
     expansion = 4
 
-    net = BNReLU(input)
     num_filters_in = net.get_shape().as_list()[1]
     # identity
-    shortcut = net
+    shortcut = input
     if stride != 1 or num_filters_in != num_filters * expansion:
-        shortcut = Conv2D('shortcut', net, num_filters * expansion, kernel_shape=1, stride=stride, use_bias=False,
+        shortcut = Conv2D('shortcut', input, num_filters * expansion, kernel_shape=1, stride=stride, use_bias=False,
                           nl=tf.identity)
     # residual
-    res = Conv2D('conv1', net, num_filters, kernel_shape=1, stride=1, use_bias=False, nl=BNReLU)
+    res = BNReLU(input)
+    res = Conv2D('conv1', res, num_filters, kernel_shape=1, stride=1, use_bias=False, nl=BNReLU)
     res = Conv2D('conv2', res, num_filters, kernel_shape=3, stride=stride, use_bias=False, nl=BNReLU)
     res = Conv2D('conv3', res, num_filters * expansion, kernel_shape=1, stride=1, use_bias=False, nl=tf.identity)
     return shortcut + res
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
@dongzhuoyao

examples/ResNet/cifar10-preact18-mixup.py
```diff
@@ -20,24 +20,86 @@
...
+    return shortcut + residual
+
+
+def bottleneck_block(input, num_filters, stride=1):
+    expansion = 4
+
+    net = BNReLU(input)
+    num_filters_in = net.get_shape().as_list()[1]
+    # identity
+    shortcut = net
```
ppwwyyxx: different from the original repo:
```python
out = F.relu(self.bn1(x))
shortcut = self.shortcut(out) if hasattr(self, 'shortcut') else x
```

ppwwyyxx:
If I'm not mistaken, the two block functions were equivalent to the ones in `resnet_model.py`  (preresnet_basicblock, preresnet_bottleneck). Better to import them than rewrite them if that is the case.

dongzhuoyao:
wx20171226-221107 2x

according to [https://github.com/kuangliu/pytorch-cifar/blob/master/models/preact_resnet.py](https://github.com/kuangliu/pytorch-cifar/blob/master/models/preact_resnet.py), the above implementation in tensorpack: line 26 and line 27 should be exchanged.  @ppwwyyxx

examples/ResNet/cifar10-preact18-mixup.py
```diff
@@ -20,24 +20,86 @@
...
+    return shortcut + res
+
+
+class ResNet_Cifar(ModelDesc):
+    # module configuration taken from reference implementation by kuangliu.github.com
+    def __init__(self, depth=18):
+        super(ResNet_Cifar, self).__init__()
+
+        if depth not in self.RESNET_CONFIG:
+            print('Could not find configuration for depth "%d". Try one of %s' % (depth, self.RESNET_CONFIG.keys()))
```
dongzhuoyao: it may be better to use logger.info("...")

yselivonchyk:
Moving BnRelu into the residual branch (the correct way) leads to diverging training. figuring out why

yselivonchyk:
@ppwwyyxx @dongzhuoyao 
Regarding your comments about design of the residual block. I ran the model by @kuangliy and figured out next correct way of applying BnRelu on residual branch: apply BnRelu before every convolution in residual branch.

The reference code expressing the ides from https://github.com/kuangliu/pytorch-cifar:
```
    def forward(self, x):
        out = F.relu(self.bn1(x))
        shortcut = self.shortcut(out) if hasattr(self, 'shortcut') else x
        ...
```
For example, for PreActResNet18 BnRelu before residual branch would be applied exactly 3 times: for the first block of the second, third and forth module (whenever filter depth changes).

I will add this changes to the pull request later.

It might also beg some changes in other Cirar+ResNet implementations as mentioned by @dongzhuoyao  but I am not that familiar with the code there to justify the changes.

dongzhuoyao:
A little clarification:

- The PreResnet have two different implementations.   the official one is [https://github.com/facebook/fb.resnet.torch/blob/master/models/preresnet.lua](https://github.com/facebook/fb.resnet.torch/blob/master/models/preresnet.lua), and the unofficial one: [https://github.com/kuangliu/pytorch-cifar/blob/master/models/preact_resnet.py](https://github.com/kuangliu/pytorch-cifar/blob/master/models/preact_resnet.py).  what the tensorpack implemented is the official PreResnet. however, the mixup paper just choosed the unofficial one.
- the unofficial one is "bn-relu first and then  diverge".
- our mixup is just an example for tensorpack, and followed the mixup authors' implementation(an unofficial preresnet).  **so I think other examples in tensorpack based on official preresnet had better keep unchanged.**
- There is one mistake that @yselivonchyk corrected for me:

> ResNet-18 with preactivation as by https://github.com/kuangliu/pytorch-cifar is using ResNet with preactivation block with 2 consecutive convolution layers in the block. Existing implementation was using 3 convolutional blocks.

- more depths choice for mixup implementations are welcome.  but just keep it only suitable for this mixup example(should not influence other examples. because I think mixup example is a special case).

yselivonchyk:
All done.
Numbers:
wd=0.0001 (reference): 6.0% no mixup 4.3% with mixup
wd=0.0005 (original)   : 5.7% no mixup 4.1% with mixup

ppwwyyxx:
https://github.com/hongyi-zhang/mixup/ FYI I found the authors have released code here.

ppwwyyxx:
Differences include:
1. how mixup is implemented (mix within 1 batch or between 2 batches)
2. initialization
3. input normalization

ppwwyyxx:
Tried the above code but found it cannot reach the paper's number either.. So maybe we can live with a slightly lower performance.

ppwwyyxx:
I can get same performance as the reference code, by using a different initializer for FC layer. Using the pytorch initializer somehow gave me worse performance starting from the first epoch and sometimes even NaN. I don't have time for now to investigate what's going on -- probably a very subtle issue. So I'll merge this one now. Thanks to @yselivonchyk and @dongzhuoyao for pointing out the bug!

Also I've removed the model code for other depth for simplicity. The same logic already exists in the resnet example so in this example it's better to just focus on mixup itself.

PR #607, opened 2018-01-24
use tf.AUTO_REUSE, add reuse arg to under_variable_scope and fix

PatWie:
plasputdata

### DIFF
--- a/tensorpack/tfutils/scope_utils.py
+++ b/tensorpack/tfutils/scope_utils.py
@@ -12,10 +12,14 @@ from ..utils.argtools import graph_memoized
 __all__ = ['auto_reuse_variable_scope', 'cached_name_scope', 'under_name_scope']
 
 
-def auto_reuse_variable_scope(func):
+def auto_reuse_variable_scope(func, per_variable=False):
     """
     A decorator which automatically reuses the current variable scope if the
     function has been called with the same variable scope before.
+
+    Args:
+        per_variable: When true, the reuse decision is made for each
+                      variable individually using `tf.AUTO_REUSE`
 
     Examples:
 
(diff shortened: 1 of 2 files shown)


tensorpack/tfutils/scope_utils.py
```diff
@@ -31,18 +31,11 @@ def myfunc(x):
...
     def wrapper(*args, **kwargs):
         scope = tf.get_variable_scope()
-        h = hash((tf.get_default_graph(), scope.name))
-        # print("Entering " + scope.name + " reuse: " + str(h in used_scope))
-        if h in used_scope:
-            with tf.variable_scope(scope, reuse=True):
-                return func(*args, **kwargs)
-        else:
-            used_scope.add(h)
+        with tf.variable_scope(scope, reuse=tf.AUTO_REUSE):
```
ppwwyyxx: It's error-prone to put any user code under `tf.AUTO_REUSE`. Because there you don't know whether your scope is reuse or not -- it's both. So a user can end up creating unexpected variables. So I prefer to keep it as it is.
Also, AUTO_REUSE is a tf1.4 thing.

PatWie: But the current auto_reuse_version has some issues as well. It affects the entire var_scope, while `tf.AUTO_REUSE` seems to be per-variable.

ppwwyyxx: By design it was meant to affect the entire scope. There is a name confusion with `tf.AUTO_REUSE` because I write this function before AUTO_REUSE appears.

ppwwyyxx: Creating unexpected (new) variables under `reuse=True` leads to an error. But it will silently pass under `tf.AUTO_REUSE`. That's the kind of error I'm talking about

tensorpack/tfutils/scope_utils.py
```diff
@@ -12,11 +12,15 @@
 __all__ = ['auto_reuse_variable_scope', 'cached_name_scope', 'under_name_scope']
 
 
-def auto_reuse_variable_scope(func):
+def auto_reuse_variable_scope(func, per_variable=False):
```
ppwwyyxx: That's not the syntax of a decorator

tensorpack/tfutils/scope_utils.py
PatWie: The `tf.AUTO_REUSE` removes workarounds like:
https://github.com/cgtuebingen/learning-blind-motion-deblurring/blob/master/learning_blind_motion_deblurring.py#L41-L47

tensorpack/tfutils/scope_utils.py
PatWie: Sorry. Then it just a one-liner pull-request

PatWie: This seems to fix the travis-ci issue.

tensorpack/tfutils/scope_utils.py
ppwwyyxx: I know `tf.AUTO_REUSE` is useful in many situations. But because it's also error-prone at the same time, I would let users to decide to use it.

ppwwyyxx: Many tf stuff (EMA, tf.layers) create variables secretly that I wouldn't have known them until I saw the error under a reuse scope.

PR #608, opened 2018-01-24
introduce some pixelart

PatWie:
I felt the readme could tolerate some colors. This is the version without the capital p.


ppwwyyxx:
The readme now looks better! I'll add that somewhere in the documentation as well.

PatWie:
What about adding a link to the model-zoo?
https://img.shields.io/badge/pre--trained-models-brightgreen.svg

pre-trained-models

PatWie:
Keras has btw a very nice start-page:
https://keras.io/

compared to

http://tensorpack.com

Is it possible to just use the readme.md as the landing page in the docs?

Issue #627, opened 2018-02-04
Allow `tf.layers` naming conventions 


ppwwyyxx:
It's not wise to keep using a different naming conventions from tf.layers despite the historical reasons. 
Both argument names and variable names should be made consistent. And we can further consider layer names.

For argument names: can just put a conversion inside each function call to translate the old names to tf.layer names, so that both can be used.
For variable names: can add an option for each layer (or maybe globally) to choose which convention to use.
Follow #291.

Argument Names:
- [x] Allow the equivalent argument names in `tf.layers` convention.
- [x] Move examples to `tf.layers` convention.
- [x] BatchNorm not done yet.
- [ ] Support more options from `tf.layers` that aren't available in tensorpack.

PR #635, opened 2018-02-07
Accept tf.layers compatible argument names

ppwwyyxx:
#627 

Both old and tf.layers argument naming should be supported now. Here are some python hacks that seems to work.

BatchNorm TBD.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -376,7 +376,7 @@ def autodoc_skip_member(app, what, name, obj, skip, options):
         'PrefetchOnGPUs',
 
         'guided_relu', 'saliency_map', 'get_scalar_var', 'psnr',
-        'prediction_incorrect', 'huber_loss',
+        'prediction_incorrect', 'huber_loss', 'SoftMax'
         ]:
         return True
     if name in ['get_data', 'size', 'reset_state']:
--- a/examples/mnist-convnet.py
+++ b/examples/mnist-convnet.py
@@ -51,9 +51,9 @@ class Model(ModelDesc):
                       .Conv2D('conv2')
                       .MaxPooling('pool1', 2)
                       .Conv2D('conv3')
-                      .FullyConnected('fc0', 512, nl=tf.nn.relu)
+                      .FullyConnected('fc0', 512, activation=tf.nn.relu)
                       .Dropout('dropout', 0.5)
-                      .FullyConnected('fc1', out_dim=10, nl=tf.identity)())
+                      .FullyConnected('fc1', 10, activation=tf.identity)())
 
         tf.nn.softmax(logits, name='prob')   # a Bx10 with probabilities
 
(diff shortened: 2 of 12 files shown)


Skylion007:
Ouch, on one hand great job for maintaining backwards compatibility. On the other hand, it's big blow to readability and documentation. ;-; Maintaining two separate keyword lists seems pretty awful and some of these hacks definitely don't follow [the principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Maybe a function decorator that would add an deprecation warning would be more readable? Perhaps something like the decorator in the top voted answer to [this](https://stackoverflow.com/questions/35567753/python-click-library-rename-argument) StackOverflow question?

ppwwyyxx:
> readability and documentation.

I expect people to read tf.layers documentation instead of mine. I could copy docs from there. I could also add a new argument list plus **kwargs.

>  the principle of least astonishment.

How does this astonish a user who doesn't go and read source code?

The click library seems to only work for command line args if I'm not mistaken. Also what I need to do is not only rename.

Skylion007:
Sorry, that Stackoverflow comment has nothing to do with the click library, but rather adding functionality that doesn't exist within it. I'll post the relevant portions.

```python
import functools

def rename_kwargs(**replacements):
    def actual_decorator(func):
        @functools.wraps(func)
        def decorated_func(*args, **kwargs):
            for internal_arg, external_arg in replacements.iteritems():
                if external_arg in kwargs:
                    kwargs[internal_arg] = kwargs.pop(external_arg)
                return func(*args, **kwargs)
        return decorated_func
    return actual_decorator


if __name__ == '__main__':

    @rename_kwargs(different_arg='format')
    def tester(different_arg):
        print different_arg

    tester(format='test value')
```
In the above snippet @rename_kwargs renames the keyword argument "format" to "different_arg" automatically. The decorator supports multiple key values passed in via a dict. You can augment the decorator to further modify the arguments as needed. That way when you decide to fully deprecate the old arguments you can just remove the decorators and the code will be readable and updated. Additionally, the decorator can be reused for all tf.layers argument translations and remove unnecessary code duplication. Isn't this how Tensorflow handles API changes anyhow?

> How does this astonish a user who doesn't go and read source code?
This is more about astonishing contributors like myself and others. It just adds pretty nasty complexity and adds to the technical debt. I definitely agree we should change the arguments, but there are better ways to do that.

Furthermore, supporting two separate versions of command arguments could get confusing. 

At the very least, do we want our examples to break whenever Tensorflow changes a default option if we rely on them for the defaults? One of the best parts of Tensorpack are the clean and well maintained examples that have excellent default values for performance.

ppwwyyxx:
Decorator looks good. I'll give it a try.

Tensorflow promises to not do breaking change so using it is fine. It's their job to not break things and I'd rather not care about it.

Skylion007:
The new changes with the decorator look good. Minor style point is that parse_args might not be the most descriptive name for what the decorator does. Otherwise, looks good.

PR #642, opened 2018-02-08
add NVML binding

PatWie:
write NVML wrapper inspired by py-nvml but written from scratch.

### DIFF
--- a/tensorpack/utils/nvml.py
+++ b/tensorpack/utils/nvml.py
@@ -3,24 +3,15 @@
 # File: nvml.py
 
 from ctypes import (byref, c_uint, c_ulonglong,
-                    CDLL, create_string_buffer,
-                    POINTER, Structure)
+                    CDLL, POINTER, Structure)
 import threading
 
 
-__all__ = ['NvidiaContext']
+__all__ = ['NVMLContext']
 
 
-_nvmlReturn_t = c_uint
-
-NVML_ERROR_LIBRARY_NOT_FOUND = 12
 NVML_ERROR_FUNCTION_NOT_FOUND = 13
-NVML_DEVICE_UUID_BUFFER_SIZE = 80
-NVML_DEVICE_NAME_BUFFER_SIZE = 64
 
-
-nvmlLib = None
-_lib_lock = threading.Lock()
 
 NvmlErrorCodes = {"0": "NVML_SUCCESS",
                   "1": "NVML_ERROR_UNINITIALIZED",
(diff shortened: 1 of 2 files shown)


PatWie:
Not sure, if you want to use a separate file `nvml.py`. Here is a stand-alone version:
https://gist.github.com/PatWie/68abc24de854ba87ff07c68dea61e778

ppwwyyxx:
A separate and __simpler__ file would be much better.

PatWie:
I think

```python
    nvidia = NvidiaContext()
    nvidia.create_context()

    num_gpus = nvidia.NumCudaDevices()

    for device in nvidia.Devices():
        print(device.Name())
        print(device.UUID())

        print(device.Memory())
        print(device.Utilization())

    nvidia.destroy_context()
```

is pretty simple. For proper usage of ctypes you need all these exception handling to not left the user with unkown errors.

tensorpack/utils/nvml.py
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# File: nvml.py
+
+from ctypes import (byref, c_uint, c_ulonglong,
+                    CDLL, create_string_buffer, NVML_ERROR_FUNCTION_NOT_FOUND,
+                    NVML_ERROR_LIBRARY_NOT_FOUND, POINTER, Structure)
```
ppwwyyxx: This import line doesn't run.

tensorpack/utils/nvml.py
```diff
@@ -0,0 +1,290 @@
...
+def CheckNvmlReturn(ret):
+    if (ret != 0):
+        raise NvmlException(ret)
+    return ret
+
+
+class NVML(object):
+
+    def load(self):
+        global nvmlLib
```
ppwwyyxx: Not necessary to make it global, right?

tensorpack/utils/nvml.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Example:
+
+            >>> print(nvidia.Device(0).UUID())
+            GPU-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+
+        Returns:
+            UUID of GPU
+        """
+        return _NVML.call(self.hnd, NVML_DEVICE_UUID_BUFFER_SIZE, "nvmlDeviceGetUUID").value
```
ppwwyyxx: I don't think UUID and name of card will ever be useful. They can be removed together with call() and BUFFER constants.

tensorpack/utils/nvml.py
```diff
@@ -0,0 +1,290 @@
...
+        super(NvidiaContext, self).__init__()
+
+    def create_context(self):
+        """Create a new context
+        """
+        _NVML.load()
+        CheckNvmlReturn(_NVML.get_function("nvmlInit_v2")())
+
+        # Atomically update refcount
+        global nvml_init_count
```
ppwwyyxx: Apart from +1 and -1 I don't see this variable used anywhere

tensorpack/utils/nvml.py
```diff
@@ -0,0 +1,290 @@
...
+        _lib_lock.acquire()
+        nvml_init_count += 1
+        _lib_lock.release()
+
+    def destroy_context(self):
```
ppwwyyxx: Make it a with statement?
` with NVMLContext() as ctx: ...`

PatWie: I thought about this as well. But for the callback version of your GPUUtilization call, you cannot use your way, as it closes and open the context for each epoch.

tensorpack/utils/nvml.py
PatWie: It think it is correct, I want a singleton-pattern here

ppwwyyxx: But everytime load() is called the so file gets loaded again. 
What's singleton is `_NVML = NVML()`. This so can just be an attribute.

ppwwyyxx: And even if it's made an attribute, is it correct to load the so everytime load() is called? Should it be cached?

PR #657, opened 2018-02-15
Printdata

PatWie:
see commit messages. The `visualize_tensors` is accidentally in this pull request. Feel free to cherry pick.

### DIFF
--- a/tensorpack/dataflow/common.py
+++ b/tensorpack/dataflow/common.py
@@ -6,6 +6,7 @@ from __future__ import division
 import numpy as np
 from copy import copy
 import pprint
+import itertools
 from termcolor import colored
 from collections import deque, defaultdict
 from six.moves import range, map
@@ -677,11 +678,11 @@ class PrintData(ProxyDataFlow):
     def __init__(self, ds, num=1, label=None, name=None, max_depth=3, max_list=3):
         """
         Args:
-            ds(DataFlow): input DataFlow.
-            num(int): number of dataflow points to print.
-            name(str, optional): name to identify this DataFlow.
-            max_depth (int, optional): stop output when to deep recursion in sub elements
-            max_list (int, optional): stop output when too sub elements
+            ds (DataFlow): input DataFlow.
+            num (int): number of dataflow points to print.
+            name (str, optional): name to identify this DataFlow.
+            max_depth (int, optional): stop output when too deep recursion in sub elements
+            max_list (int, optional): stop output when too many sub elements
         """
         super(PrintData, self).__init__(ds)
         self.num = num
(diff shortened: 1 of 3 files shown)


tensorpack/dataflow/common.py
```diff
@@ -684,61 +690,97 @@ def __init__(self, ds, num=1, label=None, name=None):
...
 
         Args:
             entry: the datapoint component
             k (int): index of this compoennt in current datapoint
             depth (int, optional): recursion depth
-
-        Todo:
-            * call this recursively and stop when depth>n for some n if an element is a list
+            max_depth (int, optional): stop output when to deep recursion in sub elements
+            max_list (int, optional): stop output when too sub elements
```
ppwwyyxx: Not used anywhere. And docs should be with \_\_init\_\_

PatWie: FixedSizeData has docs in __init__, JoinData has docs in class. This is not consistent?!

ppwwyyxx: Yes that's a bit inconsistent.. but at least arguments should be with the function.

PatWie: Just to not being confused in 61e02bc the arguments are documented? Could you somehow document how to locally build the documentation such that it can be fixed without pull-requests?

ppwwyyxx: Now are documented. I would remove the duplicated docs from here, or mark it "the same as init".
Does the readme in https://github.com/ppwwyyxx/tensorpack/tree/master/docs work?

tensorpack/dataflow/common.py
```diff
@@ -657,23 +657,31 @@ def get_data():
...
+                  ...
             datapoint 1<2 with 4 components consists of
-               dp 0: is float of shape () with range [5.88252075399]
-               dp 1: is ndarray of shape (64, 64) with range [0.0072, 0.9371]
-               dp 2: is ndarray of shape (64, 64) with range [-0.9011, 0.8491]
-               dp 3: is ndarray of shape (9, 9) with range [-0.5585, 0.5585]
+               0: float with value 5.88252075399
+               1: ndarray:int32 of shape (64,) in range [0, 10]
+               2: ndarray:float32 of shape (64, 64) with range [-0.9011, 0.8491]
+               3: dp 0: list of len 50
```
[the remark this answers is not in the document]
PatWie: Sorry, I missed that one during rebasing

tensorpack/dataflow/common.py
PatWie: Oh silly me. I didn't saw the docs dir. I tested it the documentation looks ok. I would prefer keeping the example in the class scope like PrintData currently does.

PR #707, opened 2018-03-22
Add option to write multiple tensorboard event files.

maciejjaskowski:
When uploading the event files to an object storage like GCS, one cannot append to the file. Therefore, it's much cheaper & faster if multiple event files are created as only the last one is actually uploaded.

I believe it's functionality is equivalent to tf.train.Supervisor

Tested on example mnist.

### DIFF
--- a/tensorpack/callbacks/monitor.py
+++ b/tensorpack/callbacks/monitor.py
@@ -206,7 +206,7 @@ class TFEventWriter(TrainingMonitor):
     """
     Write summaries to TensorFlow event file.
     """
-    def __init__(self, logdir=None, max_queue=10, flush_secs=120):
+    def __init__(self, logdir=None, max_queue=10, flush_secs=120, multiple_event_files=False):
         """
         Args:
             Same as in :class:`tf.summary.FileWriter`.
@@ -218,6 +218,7 @@ class TFEventWriter(TrainingMonitor):
         self._logdir = logdir
         self._max_queue = max_queue
         self._flush_secs = flush_secs
+        self._multiple_event_files = multiple_event_files
 
     def __new__(cls, logdir=None, max_queue=10, flush_secs=120):
         if logdir is None:
@@ -242,6 +243,9 @@ class TFEventWriter(TrainingMonitor):
 
     def _trigger(self):     # flush every epoch
         self._writer.flush()
+        if self._multiple_event_files:
+            self._writer.close()
+            self._writer.reopen()  # open new file
 
     def _after_train(self):
         self._writer.close()
--- a/tensorpack/callbacks/monitor.py
+++ b/tensorpack/callbacks/monitor.py
@@ -206,11 +206,13 @@ class TFEventWriter(TrainingMonitor):
     """
     Write summaries to TensorFlow event file.
     """
-    def __init__(self, logdir=None, max_queue=10, flush_secs=120, multiple_event_files=False):
+    def __init__(self, logdir=None, max_queue=10, flush_secs=120, split_files=False):
         """
         Args:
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
I think a non-appendable filesystem would be a potential problem for a lot of other things anyway. Could you just use `logdir` so that event file won't be on GCS?

Splitting event files looks very hacky. I don't think Supervisor is doing anything like that. Could you point where?

maciejjaskowski:
A non-appendable filesystem is what GCS, S3 etc. provide as object based filesystems scale much better without sacrificing resiliency ([more](https://datastudio.google.com/reporting/15RHc28q7bPfkLzZH9JjMuykqPPH7lho4/page/E2oC)).

After applying this patch, our average cost of uploading data to GCS dropped over 5 times.

As a demagogical argument: I heard from friends at Google, that's what people at Google do :-)

Indeed, I can't find specific example of such use in Supervisor. 

Still, this PR provides new feature.

ppwwyyxx:
OK. Just to clarify what you're trying to achieve: how are people expected to use this feature? Setting `logdir` to somewhere on gcs?

If yes, what happened when you set `logdir` to gcs without this feature?

maciejjaskowski:
I am not sure if setting logdir to sth on gcs works right for tensorpack now although this could be a viable option. It does work with pure tensorflow.

Assuming it works (which is likely), if you set logdir to gs:// each time a file is flushed the whole file is transferred to gcs. 

You can test it for yourself if you create a bucket YOUR_VERSIONED_BUCKET,
enable object versioning on GCS: `gsutil versioning set on YOUR_VERSIONED_BUCKET`

and use the code below:
```
import tensorflow as tf
import time

fw = tf.summary.FileWriter(YOUR_VERSIONED_BUCKED)
for i in range(5):
  value = i
  s = tf.Summary(value=[tf.Summary.Value(tag="summary_tag", simple_value=value),
])
  fw.add_summary(s)
  fw.flush()
  print("Sleeping")
  time.sleep(2)```

and then when you verify: `gsutil ls -a YOUR_VERSIONED_BUCKET` you get sth like:

```YOUR_VERSIONED_BUCKET/test-event/events.out.tfevents.1523609555.Maciejs-MBP-2.waw.nomagic.io#1523609556432534
YOUR_VERSIONED_BUCKET/test-event/events.out.tfevents.1523609555.Maciejs-MBP-2.waw.nomagic.io#1523609557845069
...```

ppwwyyxx:
If I understand it correctly, the issue is that each flush creates a new copy of the entire file due to versioning, right?
It looks like it's just because versioning doesn't make sense if you're appending to files. What if you turn versioning off?

maciejjaskowski:
Yes, each flush creates a new copy. However, it does not matter whether versioning is on or not.
I turned on versioning only to highlight the problem.

ppwwyyxx:
OK. When versioning is turned off, each flush recreates the entire file (i.e., takes unnecessary extra time but no extra space), is this correct?

Also, `tf.summary.FileWriter` itself will internally flush the file once a while. Will that also be a problem for you?

maciejjaskowski:
It does not take space but it does take money for transfer :)
Effectively, if the size (N) of a file grows linearly, the total transfer cost grows with O(N^2) in the current version of tensorpack. 
The option above lowers the cost back to O(N).

Flushing the file every now and than if no changes actually occur to the content of the file should not be a problem. I can check it if you wish. 

I was thinking, maybe I should rename the argument to sth like: rotate_logs ?

ppwwyyxx:
I see. Make sense! 

I think "rotate" means "deleting old ones while adding new ones" which is not really what's going on here. Maybe rename to "split_files" or "split_events" ?

PR #746, opened 2018-05-04
Removed excess copying when fetching data from ZMQ

tals:
The `.bytes` explicitly triggers a copy in order to create a Python `bytes` object. This isn't necessary for pyarrow deserialization.

### DIFF
--- a/tensorpack/dataflow/parallel.py
+++ b/tensorpack/dataflow/parallel.py
@@ -399,8 +399,9 @@ class MultiThreadPrefetchData(DataFlow):
 
     def __del__(self):
         for p in self.threads:
-            p.stop()
-            p.join()
+            if p.is_alive():
+                p.stop()
+                p.join()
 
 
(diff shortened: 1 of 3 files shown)


tals:
See https://github.com/zeromq/pyzmq/blob/master/zmq/backend/cython/message.pyx#L322

ppwwyyxx:
Ah yes. Before switching to pyarrow this was needed.

PatWie:
Could this please handled by the specific `load` function implementation rather than directly removing `bytes`, iff `bytes` was necessary for msgpack. Currently, new tensorpack versions would break all existing code relying on msgpack.

ppwwyyxx:
Do you have a failure case? I don't remember what the issue was.

ppwwyyxx:
Seems to work fine. Chances are it was not an issue any more for msgpack.

ppwwyyxx:
btw, TENSORPACK_SERIALIZE was added.

Issue #756, opened 2018-05-09
KerasModelCaller: "UnboundLocalError: local variable 'M' referenced before assignment"


bzamecnik:
In `tensorpack.keras.KerasModelCaller` there's a `try-finally` statement and in the `finally` block a local variable `M` representing `KerasModel` is referenced even though an exception might occur which prevented assigning it. Anyway the whole block of code seems very strange and a misuse of this language construct.

If I understand the intention, we want to return model outputs but call some procedure on the model in all cases. Why can't we just refactor it to a function and call it at the end?

PR #758, opened 2018-05-09
Refactor KerasModelCaller to prevent using an unassigned variable.

bzamecnik:
Should #756. Instead of misusing try-finally the common code is just extracted in a method and then called before returning. This should not fail on unassigned variable.


ppwwyyxx:
Makes sense!

bzamecnik:
Thanks. Unfortunately, `keras.py` is not covered by tests. I only tried that in my Jupyter notebook and went fine - it just uncovered an unrelated exception that caused the variable being unassigned.

PR #778, opened 2018-05-29
support tf.layers in argscope

PatWie:
As `tf.layers` already contains most basic operation it would be nice to extend `argscope` from Tensorpack to support `tf.layers`.

This is **not** a ready-to-merge pull-request.

### DIFF
--- a/tensorpack/tfutils/argscope.py
+++ b/tensorpack/tfutils/argscope.py
@@ -7,7 +7,7 @@ import copy
 from functools import wraps
 from inspect import isfunction, getmembers
 
-__all__ = ['argscope', 'get_arg_scope', 'enable_argscope_for_lib']
+__all__ = ['argscope', 'get_arg_scope', 'enable_argscope_for_module']
 
 _ArgScopeStack = []
 
@@ -73,14 +73,16 @@ def argscope_mapper(func):
         actual_args.update(kwargs)
         out_tensor = func(*args, **actual_args)
         return out_tensor
-    a = wrapped_func
     # argscope requires this property
-    a.symbolic_function = None
-    return a
+    wrapped_func.symbolic_function = None
+    return wrapped_func
 
 
-def enable_argscope_for_lib(module):
-    """Overwrite functions of given module to support argscope
+def enable_argscope_for_module(module):
+    """
+    Overwrite all functions of a given module to support argscope.
+    Note that this function monkey-patch the module and therefore could have unexpected consequences.
+    It has been only tested to work well with `tf.layers` module.
     """
     for name, obj in getmembers(module):
         if isfunction(obj):
--- a/tensorpack/tfutils/argscope.py
+++ b/tensorpack/tfutils/argscope.py
@@ -79,9 +79,9 @@ def argscope_mapper(func):
     return a
 
 
-def enable_argscope_for_lib(lib, decorator=argscope_mapper):
-    """Overwrite functions of given lib to support argscope
+def enable_argscope_for_lib(module):
+    """Overwrite functions of given module to support argscope
     """
-    for name, obj in getmembers(lib):
+    for name, obj in getmembers(module):
         if isfunction(obj):
-            setattr(lib, name, decorator(obj))
+            setattr(module, name, argscope_mapper(obj))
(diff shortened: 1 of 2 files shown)


PatWie:
The current state allows to write something along the lines

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

import tensorflow as tf
from tensorpack import *


patch_tflayers(tf.layers)

image = tf.placeholder(tf.float32, (None, 28, 28), 'image')
image = tf.expand_dims(image, 3) * 2 - 1

with argscope([tf.layers.conv2d], padding='same', activation=tf.identity):
    net = tf.layers.conv2d(image, 32, 3, name='conv0')

```

which will produce the output
```
('passed kwargs', {'padding': 'same', 'activation': <function identity at 0x7f6092031ed8>, 'name': 'conv0'})
```
as expected. This is just intended to discuss possible side-effects I am not aware of. Ideally, this `tf.layers` patch would further support the overview table of all used layers.

tensorpack/tfutils/argscope.py
```diff
@@ -60,3 +62,30 @@ def get_arg_scope():
...
+        actual_args.update(kwargs)
+        out_tensor = func(*args, **actual_args)
+        print('passed kwargs', actual_args)
+        return out_tensor
+    a = wrapped_func
+    a.symbolic_function = None
+    return a
+
+
+def patch_tflayers(module, decorator=tflayers_mapper):
```
ppwwyyxx: Maybe don't need to pass in the module, since this function is written for `tf.layers` only

ppwwyyxx:
Q1,2: I guess it's better to let users do this explicitly since this is a monkey-patch hack. And if this is the case, hard-coding a list may not be necessary.

Q3: I don't know what the future of these libraries would be. I think for now supporting `tf.layers` alone is good enough. 
Btw I've heard from a couple of TF people that they are encouraging people to use `tf.keras.layers` instead of `tf.layers`..

tensorpack/tfutils/argscope.py
```diff
@@ -60,3 +62,30 @@ def get_arg_scope():
...
+    """
+    @wraps(func)
+    def wrapped_func(*args, **kwargs):
+        actual_args = copy.copy(get_arg_scope()[func.__name__])
+        actual_args.update(kwargs)
+        out_tensor = func(*args, **actual_args)
+        print('passed kwargs', actual_args)
+        return out_tensor
+    a = wrapped_func
+    a.symbolic_function = None
```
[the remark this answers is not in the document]
PatWie: Well, you blame the author of [this line](https://github.com/ppwwyyxx/tensorpack/blob/099975c56b7e63d41da1da57549fedf8bd68084f/tensorpack/tfutils/argscope.py#L41). Adding `symbolic_function` here, keeps the changes minimal.

tensorpack/tfutils/argscope.py
PatWie: I renamed this function as it is quite general.

PatWie:
I changed the mnist example with `tf.layers`. The function `enable_argscope_for_lib` is be a little bit more generic as now `enable_argscope_for_lib(tf.layers)` is sufficient. I do not recommend to hard-coded `tf.layers` in `argscope.py` to not introduce another dependency.

Well, `tf.keras.layers` :unamused: is a different beast. I have an aversion to a library trying to wrap plain tensors (same reason I dislike the upper-case `tf.layers` methods). Is there any advantage besides direct access to parameters?

tensorpack/tfutils/argscope.py
```diff
@@ -60,3 +62,26 @@ def get_arg_scope():
...
+        actual_args.update(kwargs)
+        out_tensor = func(*args, **actual_args)
+        return out_tensor
+    a = wrapped_func
+    # argscope requires this property
+    a.symbolic_function = None
+    return a
+
+
+def enable_argscope_for_lib(lib, decorator=argscope_mapper):
```
ppwwyyxx: Why a `decorator` argument?
If the purpose of this function is "enable_argscope" then why would a user want to use a different decorator?

ppwwyyxx: "module" may be a better name than "lib"

PR #792, opened 2018-06-15
allow verbose shape information output for tf.layers

PatWie:
Improving the information when using `tf.layers` similar to the layers provided by tensorpack (like Conv2D, etc). I don't know if this is the perfect solution.

It produces outputs like

```
[0615 09:36:52 @argscope.py:90]                conv0:    [None, 28, 28, 1] ->   [None, 28, 28, 32]
[0615 09:36:52 @argscope.py:90]             <unkown>:   [None, 28, 28, 32] ->   [None, 14, 14, 32]
[0615 09:36:52 @argscope.py:90]                conv1:   [None, 14, 14, 32] ->   [None, 14, 14, 32]
[0615 09:36:52 @argscope.py:90]                conv2:   [None, 14, 14, 32] ->   [None, 14, 14, 32]
[0615 09:36:52 @argscope.py:90]             <unkown>:   [None, 14, 14, 32] ->     [None, 7, 7, 32]
[0615 09:36:52 @argscope.py:90]                conv3:     [None, 7, 7, 32] ->     [None, 7, 7, 32]
[0615 09:36:52 @argscope.py:90]             <unkown>:     [None, 7, 7, 32] ->         [None, 1568]
[0615 09:36:52 @argscope.py:90]                  fc0:         [None, 1568] ->          [None, 512]
[0615 09:36:52 @argscope.py:90]             <unkown>:          [None, 512] ->          [None, 512]
[0615 09:36:52 @argscope.py:90]                  fc1:          [None, 512] ->           [None, 10]
```

### DIFF
--- a/tensorpack/tfutils/argscope.py
+++ b/tensorpack/tfutils/argscope.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 import copy
 from functools import wraps
 from inspect import isfunction, getmembers
+from ..utils import logger
 
 __all__ = ['argscope', 'get_arg_scope', 'enable_argscope_for_module']
 
@@ -64,7 +65,7 @@ def get_arg_scope():
         return defaultdict(dict)
 
 
-def argscope_mapper(func):
+def argscope_mapper(func, log_shape=True):
     """Decorator for function to support argscope
     """
     @wraps(func)
@@ -72,13 +73,17 @@ def argscope_mapper(func):
         actual_args = copy.copy(get_arg_scope()[func.__name__])
         actual_args.update(kwargs)
         out_tensor = func(*args, **actual_args)
+        if log_shape:
+            in_tensor = args[0]
+            name = '<unkown>' if 'name' not in kwargs else kwargs['name']
+            logger.info('%20s: %20s -> %20s' % (name, in_tensor.shape.as_list(), out_tensor.shape.as_list()))
         return out_tensor
     # argscope requires this property
     wrapped_func.symbolic_function = None
     return wrapped_func
 
 
-def enable_argscope_for_module(module):
+def enable_argscope_for_module(module, log_shape=True):
     """
     Overwrite all functions of a given module to support argscope.
     Note that this function monkey-patches the module and therefore could have unexpected consequences.
@@ -86,4 +91,4 @@ def enable_argscope_for_module(module):
     """
     for name, obj in getmembers(module):
         if isfunction(obj):
-            setattr(module, name, argscope_mapper(obj))
+            setattr(module, name, argscope_mapper(obj, log_shape=log_shape))
--- a/tensorpack/tfutils/argscope.py
+++ b/tensorpack/tfutils/argscope.py
@@ -7,6 +7,7 @@ import copy
 from functools import wraps
 from inspect import isfunction, getmembers
 from ..utils import logger
+import tensorflow as tf
 
 __all__ = ['argscope', 'get_arg_scope', 'enable_argscope_for_module']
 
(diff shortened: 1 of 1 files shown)


tensorpack/tfutils/argscope.py
```diff
@@ -78,12 +81,46 @@ def wrapped_func(*args, **kwargs):
...
     Note that this function monkey-patches the module and therefore could have unexpected consequences.
     It has been only tested to work well with `tf.layers` module.
     """
     for name, obj in getmembers(module):
         if isfunction(obj):
-            setattr(module, name, argscope_mapper(obj))
+            setattr(module, name, func(obj))
+
+
+def enable_argscope_for_module(module):
```
ppwwyyxx: What about adding an option like `log_shape=True` here? Easier to use than a new API.

tensorpack/tfutils/argscope.py
```diff
@@ -78,12 +81,46 @@ def wrapped_func(*args, **kwargs):
     return wrapped_func
 
 
-def enable_argscope_for_module(module):
+def input_output_shape_decorator(func):
+    """Decorator for function to support input-output-shape information
+
+    Example:
```
ppwwyyxx: Let's make the format the same as tensorpack.models

PatWie: I guess
```
conv0:    [None, 28, 28, 1] ->   [None, 28, 28, 32]
```
is easier to parse than
```
conv0: input [None, 28, 28, 1]
conv0: output [None, 28, 28, 32]
```

ppwwyyxx: Agree. In tensorpack.models, layers can be nested, layers can have >1 inputs or outputs. That's why the logs are like that.
I think this format is fine.

tensorpack/tfutils/argscope.py
```diff
@@ -64,26 +65,30 @@ def get_arg_scope():
...
+def argscope_mapper(func, log_shape=True):
     """Decorator for function to support argscope
     """
     @wraps(func)
     def wrapped_func(*args, **kwargs):
         actual_args = copy.copy(get_arg_scope()[func.__name__])
         actual_args.update(kwargs)
         out_tensor = func(*args, **actual_args)
+        if log_shape:
+            in_tensor = args[0]
```
PatWie: I can add here something similar to https://github.com/tensorpack/tensorpack/blob/fa34d239ce0a6d5cc12bcb12ad690f07df5f1f0e/tensorpack/models/registry.py#L118-L119

But this would introduce a tensorflow as a dependency to this file.

ppwwyyxx: Yeah it's best to make multigpu training only print shapes in the first tower, but I'm OK without this feature. It's fine for files in `tfutils` to depend on tf.

In fact I'm not sure how variable/name_scopes in tf.layers work.. I remember it's quite complicated..

PatWie: Indeed, this is not trivial as I cannot access the scope inside `tf.layers.lowercase_name_function`.

The lastest commit in this PR at least can distinguish between the towers

```python
from tensorpack import *
import tensorflow as tf

enable_argscope_for_module(tf.layers)

def build(x):
    return tf.layers.conv2d(x, 32, [3, 3], name='conv0')
ans = []

print "----------------------------------"
plhdr = tf.placeholder(tf.float32, [1, 28, 28, 3])
ans.append(build(plhdr))

print "----------------------------------"
with TowerContext('tower0', True, vs_name='tower0') as ctx:
    ans.append(build(plhdr))

print "----------------------------------"
with TowerContext('tower1', True, vs_name='tower1') as ctx:
    ans.append(build(plhdr))

print "----------------------------------"
with TowerContext('tower2', True, vs_name='tower2') as ctx:
    ans.append(build(plhdr))


```

PatWie: The `InferenceTower` is still dumped to the console. Which is ok, as it might differ from the training tower.

PR #794, opened 2018-06-18
Dataflow ZMQ paralelization for nonrandom sequences

Darthholi:
Hello! 
We love dataflow and actually need to paralelize a specific generator, that needs to run over a list of items. That, of course, produced duplicates, so we came up with a solution.

Please, when You will have time, can You look at the changes? I tried to make the changes as minimal as they could be (the other way to add a slicing behavior would be to use reset_state for the generators, but that seemed to me like a lot of changes to existing code, so instead I tried to use new classes).

Thank You very much!

### DIFF
--- a/tensorpack/dataflow/base.py
+++ b/tensorpack/dataflow/base.py
@@ -190,8 +190,8 @@ class DataFlowSequenceSlicer(ProxyDataFlowSequence):
         if None in [self.slicing_n, self.slicing_i]:
             return self.ds.size()
         else:
-            remainder_add = 1 if (self.ds.size() % self.slicing_n >= self.slicing_i) else 0
-            return int(self.ds.size() / self.slicing_n) + remainder_add
+            remainder_add = 1 if (self.ds.size() % self.slicing_n > self.slicing_i) else 0
+            return int(self.ds.size() // self.slicing_n) + remainder_add
 
 def is_sequence(ds):
     return isinstance(ds, DataFlowSequence) or (hasattr(ds, '__getitem__') and hasattr(ds, '__len__'))
--- a/tensorpack/dataflow/base.py
+++ b/tensorpack/dataflow/base.py
@@ -78,10 +78,44 @@ class DataFlow(object):
         """
         pass
 
+class DataFlowSequence(DataFlow):
+    """
+    A DataFlow based on a sequence. Can be sliced then.
+    Should be similar in look to https://github.com/keras-team/keras/blob/master/keras/utils/data_utils.py class Sequence
+    """
+
+    def get_data(self):
+        """Create an infinite generator that iterate over the Sequence."""
+        while True:
+            for item in (self[i] for i in range(len(self))):
+                yield item
+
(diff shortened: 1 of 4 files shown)


Darthholi:
(And if You know how to make the checks pass, tell me, i would love to fix it of course! :) )

ppwwyyxx:
These are too many changes and it effectively adds a new interface of DataFlow, and you are on the road towards reimplementing this new interface for __every__ existing DataFlow..

There are simpler solutions for your use case. See https://github.com/tensorpack/tensorpack/issues/401#issuecomment-328963606.
You just need to write one "base" dataflow that produces different set of indices in each process.

Darthholi:
Thanks! 
I will just ask, for clarity - the base dataflow producing different set of indices needs to know to which process it belongs and the total number of processes.
Where, in the code, would You prefer to give it the information?

My take would be to use reset_data(adding the slicing params here) called from ZMQ that would propagate to the base dataflow.
(I did write the new classes to avoid cluttering reset_data with params, because I was afraid You would not like the solution, but maybe I was wrong :) )

bzamecnik:
I agree with @ppwwyyxx that this is a big change to the existing object model.

So far a forking operation just clones the given data flow graph and runs multiple instances of each.

What we'd like to achieve is to fork a dataflow into multiple parallel map operations, each working on a different slice. I can imagine slicing on infinite generators (done by a cooperative way by predefined rule) - keep sample if `index % total_workers == worker_index`. For that to work efficiently the pipeline would need to be evaluated in a lazy way, so that we'd skip computation for skipped samples. This assumes that complexity of mapping function is uniform. Otherwise we'd need some queue to balance the computation.

Anyway I'm starting to think whether Dask with it's task graph being a full DAG would be more suitable for our input pipeline.

ppwwyyxx:
I was a bit misleading. The solution in https://github.com/tensorpack/tensorpack/issues/401#issuecomment-328963606 does not guarantee strictly non-duplication. It could have a small amount of duplication at the end of the epoch, because those data could possibly mix with the next epoch.

At that time there was no `MultiProcessMapData`. What I had in mind is to have a single dataflow that produces all indices (so it's fast enough and don't need parallelization), and then run the function that maps indices to data in parallel, such as:

```python
ds = DataFlowProducesIndices()

def mapper(i):
    return cv2.imread(filenames[i])

ds = MultiProcessMapData(ds, 10, mapper, strict=True)
```

I think this can achieve what you expect.

Darthholi:
Thank You!

We had a conversation with @bzamecnik and we would like to ask You how is it achieved without giving the generator the information about the processID (and total number of processes)...?

Because if, in the end, this information would be needed, Iam ready to code it for us all.

ppwwyyxx:
`ds = DataFlowProducesIndices()` will only exist in one process. It will not be forked.
`MultiProcessMapData` creates processes to only run the mapper function.

ppwwyyxx:
Closing as it looks like the problem was solved.

PR #795, opened 2018-06-18
Allow CacheData to store data on disk

PatWie:
In some tasks a fair comparison requires the exact same validation
set. While this could be done manually by storing into LMDB and
reading from LMDB it is cumbersome to do so. Instead this commit
allows to specify a storage file in `CacheData` which stores the
incoming dataflow. Whenever the same dataflow is used it will
produce exactly the same dataflow. This is important when the
validation set was generated in a synthetic way, see below.

Example (denoising):

```python
from tensorpack import *

ds = dataset.Mnist('test')
ds = FixedSizeData(ds, 5)
ds = MapData(ds, lambda dp: [dp[0], dp[0]])
ds = AugmentImageComponent(ds, [imgaug.GaussianNoise(0.1)])
ds = PrintData(ds)
ds = CacheData(ds, storage_file='mnist_test.npz')
ds.reset_state()
(potentially different models) runs.
```

Without `CacheData(ds, storage_file='mnist_test.npz')` the
validation set changes between different runs (not epochs!) making
any comparisons of architectural changes impossible.

### DIFF
--- a/tensorpack/dataflow/common.py
+++ b/tensorpack/dataflow/common.py
@@ -11,6 +11,7 @@ from termcolor import colored
 from collections import deque, defaultdict
 from six.moves import range, map
 import tqdm
+import os
 
 from .base import DataFlow, ProxyDataFlow, RNGDataFlow, DataFlowReentrantGuard
 from ..utils import logger
@@ -610,13 +611,16 @@ class CacheData(ProxyDataFlow):
     NOTE: The user should not stop the iterator before it has reached the end.
         Otherwise the cache may be incomplete.
     """
-    def __init__(self, ds, shuffle=False):
+    def __init__(self, ds, shuffle=False, storage_file=None):
         """
         Args:
             ds (DataFlow): input DataFlow.
             shuffle (bool): whether to shuffle the datapoints before producing them.
+            storage_file (string): file which stores the cache for re-using (potentially randomly augmented) data
+                                   between different runs
         """
         self.shuffle = shuffle
+        self.storage_file = storage_file
         self._guard = DataFlowReentrantGuard()
         super(CacheData, self).__init__(ds)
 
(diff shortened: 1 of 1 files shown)


PatWie:
I know the `npz` can be quite large. A solution is to use two different `InferenceRunner` (small with CacheData and large with generated on-the-fly data).

ppwwyyxx:
Why not a function that dumps a dataflow to npz, and a dataflow that reads from npz? Like `DataFlowFromNpz(filename), DataFlowFromNpz.dump(ds, filename)`. This way we add new stuff but it's easier to understand and composable. Also allows other formats this way. 

And you can then move the logic of `if file exists:   ... else: ...` to your code. Having to explain an if-else logic in docs may suggest that the function may be doing too many things.

PatWie:
I was hoping for such a discussion. It not clear to me, what the solution would be. In fact, I don't even think npz is the best way?

I was thinking about something along the lines of:
```
ds = CacheData(ds, serialize_obj=NpzSerializer('cache.npz'))
```

I just put it in the current draft into "CacheData", since otherwise, I don't need CacheData at all. But using `DataFlowFromNpz` and `DataFlowFromNpz` would put the if-else logic into the user-code and needs to be repeated all the time. That's what I am currently doing all the time.
I just want to prevent me from doing the same stuff again and again.

The problem or discussion should be: How can we ensure, that the validation data which *can* be generated on-the-fly is stored in some persistent way for a fair comparison of different models.

My current way is:
- writing a `data_sampler.py` storing stuff into `train.lmdb` and `val.lmdb`.
- writing a `data_provider.py` doing stuff on-the-fly
- running `data_provider.py` for just the validation data and dump ot again into `val-augmentation-baked-in.lmdb`

- train network using  'train.lmdb`  and `val-augmentation-baked-in.lmdb`. Any suggestions how to reduce the code and not to just rewrite it?

ppwwyyxx:
> the validation data which can be generated on-the-fly is stored in some persistent way for a fair comparison of different models.

Since this is the main goal, I think it naturally expects a logic to tell "do I need to generate data" or "do I need to load data". Check the existence of file is one way to do that and I'd prefer the user to manage such high-level logic on "what to do".

To reduce code, what about the following?:
```python
def get_val_data():
    if not exists(file):
        ds = ....
        ds = CacheData(ds)
        DataFlowFromNpz.dump(ds, file)
    return DataFlowFromNpz(file)
```

PatWie:
closed in favor of #797

PR #797, opened 2018-06-19
Refactor serialize

PatWie:
I felt the serialize and load methods need some refactoring there were some issues:

- dump_dataflow_to_tfrecord had a missing assert
- dump_dataflow_to_lmdb had a bug wrt to the ordering
- dump_dataflow_to_lmdb and dump_dataflow_to_tfrecord share a lot of functionality and code duplication was there. I would be a bad idea to add yet another dump_dataflow_to_<something>

In this commit, I simplified the usage, by having counter-parts

LMDBDataWriter <-> LMDBDataReader
NumpyDataWriter <-> NumpyDataReader
HDF5DataWriter <-> HDF5DataReader
TFRecordDataWriter <-> TFRecordDataReader

with will intuitive names.

[Example is here](https://github.com/PatWie/tensorpack-fork/blob/f5f06ef8296d82a2d5f9e587bcae5dfc0a55111e/tests/test_datawriter.py#L45-L81)

The old LMDBDataPoint, TfRecordData are still working having a deprecation warning. I am planning to add the deprecation warning as well to dump_dataflow_to_tfrecord and dump_dataflow_to_lmdb as they just reusing TfRecordDataWriter and LMDBDataWriter.

## Unit Tests
Importantly!, there are now unit tests for these read- and write-options. These indicates, there there was an issue with LMDBDataPoint and the old `dump_dataflow_to_lmdb` as well as the new `LMDBDataWriter` implementation. The latter is now fixed and `dump_dataflow_to_lmdb` is just a wrapper for the new implementation for backward compatibility. So now it works. Travis-CI is a different story.

## Example

```python
# writing some data
ds = SomeData()
NumpyDataWriter(ds, 'mydataset.npz').serialize()
# loading some data
ds2 = NumpyDataReader('mydataset.npz')
```

and replace `NumpyData???` with `LMDBData???`, `HDF5Data???` , `TFRecordData???`  to use another format.


PatWie:
The issue in the old `dump_dataflow_to_lmdb_old` has been

```python
txn.put(u'{}'.format(idx).encode('ascii'), dumps(dp))
```
instead of 
```python
txn.put(u'{:08}'.format(idx).encode('ascii'), dumps(dp))
```

see docs:

> LMDB is a tiny database with some excellent properties:
> - Ordered map interface (keys are always lexicographically sorted).

PatWie:
When no changes are necessary, I will update the docs.

PatWie:
Travis is failing due to some pyarrow issues. I do not use pyarrow. A workaround is checking 
```python
if os.environ.get('CONTINUOUS_INTEGRATION', 'false') == 'false':
   do_test() # we are not in travis and can do the test
``` 
(but this smells very bad).

Here [`CONTINUOUS_INTEGRATION` is a default ENV-var](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables)

PatWie:
The fix is working, but benchmarking shows some bad results:

```python
import msgpack
import pyarrow as pa
import time

import numpy as np

buf = []
for _ in range(1000):
    buf.append(np.random.randn(256, 256, 3))


elapsed = time.time()
s1 = 0
for dp in buf:
    enc = pa.serialize(dp).to_buffer().to_pybytes()
    el = pa.deserialize(enc)
    s1 += el.sum()
elapsed = time.time() - elapsed
print('pyarrow.to_buffer().to_pybytes()', elapsed)

elapsed = time.time()
s2 = 0
for dp in buf:
    enc = pa.serialize(dp).to_buffer()
    el = pa.deserialize(enc)
    s2 += el.sum()
elapsed = time.time() - elapsed
print('pyarrow.to_buffer()', elapsed)

# elapsed = time.time()
# s3 = 0
# for dp in buf:
#     enc = msgpack.dumps(dp, use_bin_type=True)
#     el = msgpack.loads(enc, raw=False)
#     s3 += el.sum()
# elapsed = time.time() - elapsed
# print('msgpack.dumps()', elapsed)
```

gives
```
('pyarrow.to_buffer().to_pybytes()', 0.4506850242614746)
('pyarrow.to_buffer()', 0.30463695526123047)
```

ppwwyyxx:
For pyarrow dumps() does not return bytes. It returns a buffer that works like bytes most of the time.

There are many changes that I'll take time to digest but I like the direction to organize readers and writers. For now I don't think it's worth increasing the complexity of interfaces just to share a progress bar among writers. This for example makes the write_frequency option in lmdb become a bit unnatural to implement.
I still prefer using functions for writers (i.e. `dump_lmdb(ds, filename)` instead of `LMDBWriter(ds, filename).write()`) because I think you'll never need to use an instance of writer twice. So you'll always call it with one line of `LMDBWriter(ds, filename).write()` which can be achieved with a function call.

PatWie:
> It returns a buffer that works like bytes most of the time.

Then I found exactly the situation where this does not hold.

I do not think this increases the complexity. All these writers are following the same interface. It is a matter of consistency: Adding `dump_npy` and `dump_hdf5` just feels wrong.

Regarding the naming conventions: I am not strong about the exact naming, I just followed `TestDataSpeed`, which could be `test_data_speed(ds)` as well for the same argument.

For sure, `dump_lmdb` and `LMDBDataPoint` is unintuitive when compared to `LMDBDataWriter` and `LMDBDataReader`. And I don't think `read_lmdb(path)` like `dump_lmdb` is a good choice either.

ppwwyyxx:
About interface:
The concept of writer exists in many libraries because there you can create a writer and call `write(bytes)` multiple times with different data -- that actually corresponds to "puts" here. That's why I was uncomfortable with the interface. Also, writers can't share the constructor because writers can have their own options and begin() can totally be put in constructor as well. 

To match common terminology of writer, the shared public interface would be `.write(datapoint)` and `.close()` and possibly also a `.dump(dataflow)` which closes itself in the end. 

About function vs class:
It looks nice when they are sharing interfaces but I think `LMDBWriter(filename).dump(dataflow)` may not look as simple as a single function call like `LMDBWriter.dump(filename, dataflow)` or even `LMDBReader.dump(filename, dataflow)`.

ppwwyyxx:
About names:
`LMDBDataReader` is not a better name than `LMDBDataPoint` because this simpler name sounds too general and may become misleading given that the class does not read __any__ LMDB. `LMDBData`, `LMDBDataDecoder` and `LMDBDataPoint` are already a bit confusing but I'm afraid that `LMDBDataReader` makes it worse. Same for `NumpyDataReader`, and I should add a `decoder=` option in `TFRecordData`. 

I think the names should at least convey the message that it reads some tensorpack-specific thing. But I also like the correspondence between `XXReader` and `XXWriter` because the reader can ONLY reads something dumped by the writer.

Taking a step back and look at the whole problem, these functions are just doing serialization-deserialization of dataflow. The traditional way of doing serialization/deserialization in python is to have `.load` and `.dump` (or  `getstate/setstate`, `tostring/fromstring`) defined in the class and users are never expected to understand the format at all. __The only difference__ here is that DataFlow needs to be serialized to multiple formats.

Going from the traditional way to creating one class for each serializer/deserializer is too far a step. I've been talking without giving any actual suggestions. But now I would propose the following interface:
```python
DataFlowSerializer.to_numpy(ds, filename)
DataFlowSerializer.from_numpy(filename)
DataFlowSerializer.to_lmdb(ds, filename, options)
....
```

or 
```python
NumpySerializer.dump(ds, filename)
NumpySerializer.load(filename)
LMDBSerializer.dump(ds, filename, options)
...
```

The benefits are:
1. They are more tightly paired (inside one namespace now, just like the traditional dump/load).
2. It's clear that they are serializers, i.e. you cannot use a LMDB deserializer to load an arbitrary lmdb file.

This is only about designing the interface and names. Implementation-wise the progress bar may still be shared.

PatWie:
Then
ˋˋˋpython
DataflowSerializer.save(ds, "db.lmdb") # same as np.save
DataflowSerializer.load("db.lmdb")
ˋˋˋ
would be most straightforward depending on the actual file extension ("lmdb,npz,tfrecord,h5") or eventually use `.save(ds, filename, format="lmdb", allow_overwrite=False)`.
I do not want to rewrite it again and again.
Let me know this sounds ok for you. 

There is one issue though: hdf5 requires a datapath argument. Tightly couple all functionalities into a single class/function look a bit weird. But I guess for the first version it would be ok.

ppwwyyxx:
Different formats may support different options (like write_frequency in lmdb and the data path for hdf5) so it's probably impossible to share the same function and dispatch based on the filename.

PatWie:
As I am absolute convince a function-interface for serialization is not an improvement, I implemented your second version:

```python
ds = SomeData()
NumpyDataSerializer('mydataset.npz').serialize(ds)
ds2 = NumpyDataSerializer('mydataset.npz').load()
```

But the implementation is really awkward. There is not so much which can be shared between `<Some>DataSerializer.serialize()` and `<Some>DataSerializer.load()`.
Currently, the `load` method simply returns the `<Some>DataReader()` object. 

Writing a dataflow to disk and reading a dataflow from disk are two very different things and require different objects. In Python preferring classes over function keeps the implementation easy and avoids code duplication. 

The more time spending about a good design, the more I prefer the my previous version from a1892b5fe0606437bd0a5f9d83d7c7f2f6239628. I suggest to do a rollback to this commit.

In this commit, there is **no** code duplication and no nesting of different objects, just

```python
ds = SomeData()
NumpyDataWriter(ds, 'mydataset.npz').serialize()
# I can live with ...
NumpyDataWriter('mydataset.npz').serialize(ds)
# ... as well, which might support multiple incoming ds in the future
# loading some data
ds2 = NumpyDataReader('mydataset.npz')
```
This look pretty straightforward and it is something I can easily remember without having a look at the tensorpack docs (a very rare thing compare to TestDataSpeed, dump_to_something).

> Going from the traditional way to creating one class for each serializer/deserializer is too far a step

I think it is eactly the opposite. Think of a base-class which provides logics to prevent overriding files, verifying dataflows.

Rollback to a1892b5fe0606437bd0a5f9d83d7c7f2f6239628?

ppwwyyxx:
There is not much to be shared among them -- and that's what I expected. There is not much duplication among them. 
Interface-wise they are different: for save, they have different options. Even "filename" is not a universal argument because LMDB can take a directory. And `overwrite=True` should be the user's responsibility. Would you call it a good design for `np.save(filename, overwrite=True)` or `pickle.dump(obj, file, overwrite=True)`? For load, they have different options (e.g. shuffle).
Implementation-wise the only code duplication is setting up the dataflow and progress bar, which is about 6 lines.
See #802. It's WIP but I think this design is much simpler.

PatWie:
I do not see any gain in having a `load` function which simply wraps 

```diff
 +    def load(path, size=None):
 +        return TFRecordData(path, size, decoder=loads)
```

over separate `TFRecordWriter` and `TFRecordData`

ppwwyyxx:
I don't mind moving the implementation of `TFRecordData` into `load()`.

The only benefit is that `TFRecordData` is for reading data with a user-defined decoder. i.e., we have one API that reads an arbitrary tfrecord, and one API that's a dataflow deserializer, and they share code. Here they are named `TFRecordData` and `TFRecordSerializer.load` respectively.

ppwwyyxx:
@PatWie do you have other suggestions on #802?

PatWie:
I only had a comment about the keys for LMDB.

PR #802, opened 2018-06-25
DataFlow Serializers

ppwwyyxx:
- [x] Change docs, examples

### DIFF
--- a/tensorpack/dataflow/serialize.py
+++ b/tensorpack/dataflow/serialize.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+# File: serialize.py
+
+import os
+import numpy as np
+
+from ..utils.utils import get_tqdm
+from ..utils import logger
+from ..utils.serialize import dumps, loads
+
+from .base import DataFlow
+from .format import LMDBData, TFRecordData
+from .common import MapData
+from .raw import DataFromList
+
+__all__ = ['LMDBSerializer', 'NumpySerializer', 'TFRecordSerializer']
+
+
+def _reset_df_and_get_size(df):
+    df.reset_state()
+    try:
+        sz = df.size()
+    except NotImplementedError:
+        sz = 0
+    return sz
+
+
+class LMDBSerializer():
+    @staticmethod
+    def save(df, path, write_frequency=5000):
+        """
+        Dump a Dataflow to a lmdb database, where the keys are indices and values
+        are serialized datapoints.
+        The output database can be read directly by
+        :class:`tensorpack.dataflow.LMDBDataPoint`.
+
+        Args:
+            df (DataFlow): the DataFlow to dump.
+            path (str): output path. Either a directory or a mdb file.
+            write_frequency (int): the frequency to write back data to disk.
+        """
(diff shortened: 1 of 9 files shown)


tensorpack/dataflow/serialize.py
```diff
@@ -0,0 +1,124 @@
...
+            txn = db.begin(write=True)
+            for idx, dp in enumerate(df.get_data()):
+                txn.put(u'{}'.format(idx).encode('ascii'), dumps(dp))
+                pbar.update()
+                if (idx + 1) % write_frequency == 0:
+                    txn.commit()
+                    txn = db.begin(write=True)
+            txn.commit()
+
+            keys = [u'{}'.format(k).encode('ascii') for k in range(idx + 1)]
```
PatWie: Should be `keys = [u'{:08}'.format(k).encode('ascii') for k in range(self.latest_idx + 1)]`

tensorpack/dataflow/serialize.py
```diff
@@ -0,0 +1,124 @@
...
+        db.close()
+
+    @staticmethod
+    def load(path, shuffle=True):
+        """
+        Args:
+            path (str): path to lmdb file
+        """
+        df = LMDBData(path, shuffle=shuffle)
+        return MapData(df, lambda dp: loads(dp[1]))
```
PatWie: Aren't these two lines exactly the `LMDBDataPoint` implementation? Why not using `LMDBDataPoint` directly here?

PatWie:
I would adapt and add the unit tests later.

tensorpack/dataflow/serialize.py
ppwwyyxx: Because the only valid use case for `LMDBDataPoint` is to use with `LMDBSerializer.save` (or previously dump_dataflow_to_lmdb), and `LMDBDataPoint` has a worse name than `LMDBSerializer.load` so should be deprecated.

`TFRecordData` is a bit different because it has other valid use cases. It can be used to load with custom decoder, at least in this PR __for now__ -- I think that's a very very rare use case anyway, so we maybe just don't support it.

PatWie: That sounds ok, I would really prefer then `LMDBSerializer.load`.

PatWie:
Not if you ask me. But it is overly commonly used.

tensorpack/dataflow/serialize.py
```diff
@@ -0,0 +1,191 @@
...
+    def save(df, path):
+        """
+        Args:
+            df (DataFlow): the DataFlow to serialize.
+            path (str): output tfrecord file.
+        """
+        size = _reset_df_and_get_size(df)
+        with tf.python_io.TFRecordWriter(path) as writer, get_tqdm(total=size) as pbar:
+            for dp in df.get_data():
+                writer.write(dumps(dp).to_pybytes())
```
PatWie: This breaks legacy implementations based on `msgpack`. Is `msgpack` still supported?

ppwwyyxx: eventually no. But I should keep msgpack working.

PR #830, opened 2018-07-16
Add unit test for Serializer

PatWie:
Test new serialize interface and HDF5 support. Mainly the missing features from #797 .

### DIFF
--- a/tensorpack/dataflow/serialize.py
+++ b/tensorpack/dataflow/serialize.py
@@ -89,6 +89,7 @@ class NumpySerializer():
     Note that this would have to store the entire dataflow in memory,
     and is also >10x slower than the other serializers.
     """
+
     @staticmethod
     def save(df, path):
         """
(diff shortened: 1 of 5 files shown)


tensorpack/dataflow/serialize.py
```diff
@@ -100,14 +100,15 @@ def save(df, path):
         size = _reset_df_and_get_size(df)
         with get_tqdm(total=size) as pbar:
             for dp in df.get_data():
-                buffer.append(dp)
+                buffer.append(dumps(dp).to_pybytes())
```
ppwwyyxx: why?

ppwwyyxx: This can be fixed by:
```
        np.savez_compressed(path, buffer=np.asarray(buffer, dtype=np.object))
```

tensorpack/dataflow/serialize.py
```diff
@@ -152,6 +154,46 @@ def load(path, size=None):
...
+    Note that HDF5 files are in fact not very performant and currently do not support lazy loading.
+    It's better to use :class:`LMDBSerializer`.
+    """
+    @staticmethod
+    def save(df, path, data_paths=[]):
+        """
+        Args:
+            df (DataFlow): the DataFlow to serialize.
+            path (str): output hdf5 file.
+            data_paths (list, optional): list of h5 paths to zipped.
```
ppwwyyxx: why is it optional and has a default value? from the code it looks like it is necessary and has to have the same length as datapoints.

tensorpack/dataflow/serialize.py
```diff
@@ -152,6 +154,46 @@ def load(path, size=None):
...
+        """
+        size = _reset_df_and_get_size(df)
+        buffer = dict()
+
+        for data_path in data_paths:
+            buffer[data_path] = []
```
ppwwyyxx: `from collections import defaultdict` ?

tensorpack/dataflow/serialize.py
PatWie: Good catch!

Issue #291, opened 2017-06-05
Provide wrappers around tf.layers


ppwwyyxx:
Since `tf.layers` is in tf core they'll hopefully be well maintained. It's easier to just call them instead of rewriting the layers.
Would also solve #289, #283, #236.
Before this feature is added, calling `tf.layers` directly would also work.

The only issue now is the name difference. Tensorpack has been using different argument names and variable names.

We could deprecate the old names, or add some global/layer-specific option to switch between the two.
Deprecation of variable names is probably not an option because models are hard to fix.

PatWie:
Does `tf.layers` follows a common parameter pattern (name, input, param1, param2)? Does their BatchNorm provides "use_local_stats"?

I also thought about a "@tflayer_register()", which is probably the smoothest transition towards `tf.layers`. Hopefully, the output of "incoming-shape, outgoing shape, the number of params, size" will still exist in the future tensorpack versions.

ppwwyyxx:
Yes I hope to keep the shape output. Just want to use tf.layers for the actual implementation.

I guess first step we can switch the implementation but keep all the APIs/names unchanged (by custom_getter).

Skylion007:
We can also try autorefactor tools to update all the examples and put deprecation warning under the old layers. Python [rope](https://github.com/python-rope/rope/blob/master/docs/rope.rst) seems like the best solution if we do that approach.

Skylion007:
Any update on this front? The longer we wait to fix this, the more troublesome an issue it will become.

ppwwyyxx:
The change is supposed to be transparent to users. There won't be any trouble as said in the issue.

Skylion007:
The question then begins if we create new layers should we support the Tensorflow arguments or the legacy Tensorpack ones? How can we easily mix the argscopes between Tensorpack and Tensorflow? Or is that what the custom getter is trying to achieve? And if Tensorpack layers are not being deprecated it becomes confusing when to use which one. Should the user register a new tf.layer or a Tensorpack layers? This seems like it is going to cause a lot of issues down the line if we don't eventually merge or deprecate the Tensorpack layers.

ppwwyyxx:
You use whichever you like, and implement new layers in whichever ways you like. They are just doing the same thing with different interface (different function name, argument name, etc) and you can always use one in place of anther. Why would there be confusion?

ppwwyyxx:
Closing. Only two layers were not switched to `tf.layers` implementation:
[...]

PR #832, opened 2018-07-18
Change VGG19 example to tf.layers.

PatWie:
As part of the transition towards tf.layers #291, #627 this is a
rewrite. While The LinearWrap might need less lines, tf.layers are
supported by TensorFlow. And the motivation of tensorpack is being
a training framework.

### DIFF
--- a/examples/CaffeModels/load-vgg19.py
+++ b/examples/CaffeModels/load-vgg19.py
@@ -7,53 +7,58 @@ import cv2
 import tensorflow as tf
 import numpy as np
 import os
+import six
 import argparse
 
 from tensorpack import *
-from tensorpack.tfutils.symbolic_functions import *
-from tensorpack.tfutils.summary import *
 from tensorpack.dataflow.dataset import ILSVRCMeta
+
+enable_argscope_for_module(tf.layers)
 
 
 def tower_func(image):
-    with argscope(Conv2D, kernel_size=3, activation=tf.nn.relu):
-        logits = (LinearWrap(image)
-                  .Conv2D('conv1_1', 64)
-                  .Conv2D('conv1_2', 64)
-                  .MaxPooling('pool1', 2)
-                  # 112
-                  .Conv2D('conv2_1', 128)
-                  .Conv2D('conv2_2', 128)
-                  .MaxPooling('pool2', 2)
-                  # 56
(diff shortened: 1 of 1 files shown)


examples/CaffeModels/load-vgg19.py
```diff
@@ -7,53 +7,58 @@
...
-                  .FullyConnected('fc8', 1000)())
+    with argscope([tf.layers.conv2d], kernel_size=3, activation=tf.nn.relu, padding='same'):
+        x = image
+        x = tf.layers.conv2d(x, 64, name='conv1_1')
+        x = tf.layers.conv2d(x, 64, name='conv1_2')
+        x = tf.layers.max_pooling2d(x, 2, 2, name='pool1')
+
+        x = tf.layers.conv2d(x, 128, name='conv2_1')
+        x = tf.layers.conv2d(x, 128, name='conv2_2')
+        x = tf.layers.max_pooling2d(x, 2, 2, name='pool1')
```
ppwwyyxx: should better be 'pool2'

examples/CaffeModels/load-vgg19.py
```diff
@@ -7,53 +7,58 @@
...
+
+        x = tf.layers.conv2d(x, 512, name='conv5_1')
+        x = tf.layers.conv2d(x, 512, name='conv5_2')
+        x = tf.layers.conv2d(x, 512, name='conv5_3')
+        x = tf.layers.conv2d(x, 512, name='conv5_4')
+        x = tf.layers.max_pooling2d(x, 2, 2, name='pool1')
+        x = tf.layers.flatten(x, name='flatten')
+
+        x = tf.layers.dense(x, 4096, activation=tf.nn.relu, name='fc6')
+        x = tf.layers.dropout(x, rate=0.5, name='drop0')
```
ppwwyyxx: This will actually do the dropout? Tensorpack's dropout layer knows the context is inference, and will not perform the dropout.

PatWie: I always test the implementation before doing a PR. The default is inference-mode for `tf.layers.dropout` (training=False).

examples/CaffeModels/load-vgg19.py
PatWie: You are right! Done in the next commit.

PR #853, opened 2018-08-04
add flownet2 inference examples

PatWie:
TODO:

- [x] FlowNet2-S
- [x] FlowNet2-C
- [x] FlowNet2


PatWie:
I know, this is only the inference part (like the VGG, AlexNet examples). I would update the links to the ported Caffe-Model if there is nothing to change. I consider to re-implement the FlowNet2 or PWC. This is a good start to do a verifiable implementation for TensorFlow.

ppwwyyxx:
I'm missing some contexts on what you're doing. I downloaded `flownet2-models.tar.gz` at https://github.com/lmb-freiburg/flownet2, and saw the following:
```
FlowNet2/    FlowNet2-C/   FlowNet2-CS/   FlowNet2-CSS/        FlowNet2-CSS-ft-sd/  FlowNet2-S/   FlowNet2-ss/  FlowNet2-sss/
FlowNet2-c/  FlowNet2-cs/  FlowNet2-css/  FlowNet2-css-ft-sd/  FlowNet2-s/          FlowNet2-SD/  FlowNet2-SS/
```
Did you port the weights of `FlowNet2-s`?

examples/OpticalFlow/flownet2s.py
```diff
@@ -0,0 +1,224 @@
...
+            concat3 = tf.concat([conv3, x, flow4_up], axis=1)
+            flow3 = tf.layers.conv2d(pad(concat3, 1), 2, name='predict_flow3', strides=1, activation=tf.identity)
+            flow3_up = tf.layers.conv2d_transpose(flow3, 2, name='upsampled_flow3_to_2', use_bias=False)
+            x = tf.layers.conv2d_transpose(concat3, 64, name='deconv2', activation=lambda x: tf.nn.leaky_relu(x, 0.1))
+
+            concat2 = tf.concat([conv2, x, flow3_up], axis=1)
+            flow2 = tf.layers.conv2d(pad(concat2, 1), 2, name='predict_flow2', strides=1, activation=tf.identity)
+
+        flow2 = tf.identity(flow2, name='flow2')
+        return resize(flow2 / 20.)
```
ppwwyyxx: What is this /20 ???

examples/OpticalFlow/flownet2s.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def build_graph(self, left, right):
+        prediction = self.graph_structure(left, right)
+        tf.identity(prediction, name="prediction")
+
+    def optimizer(self):
```
ppwwyyxx: This part does not make sense for now. Can be removed.

examples/OpticalFlow/flownet2s.py
```diff
@@ -0,0 +1,224 @@
...
+class FlowNet2S(ModelDesc):
+    def inputs(self):
+        return [tf.placeholder(tf.float32, (None, CHANNELS, SHAPE, SHAPE), 'left'),
+                tf.placeholder(tf.float32, (None, CHANNELS, SHAPE, SHAPE), 'right')]
+
+    def graph_structure(self, left, right):
+        # this is refactored into a separate function for a potential re-implementation of the entire FlowNet2
+        x = tf.stack([left, right], axis=2)
+        rgb_mean = tf.reduce_mean(x, axis=[0, 2, 3, 4], keep_dims=True)
+        x = (x - rgb_mean) / 255.
```
ppwwyyxx: Perhaps it does not matter much, but is this actually what they're doing? It seems to be some custom caffe code so I didn't follow much into it.

examples/OpticalFlow/flownet2s.py
PatWie: This is an implementation detail. The network is trained on scaled flow-field (by factor 20.)

examples/OpticalFlow/flownet2s.py
PatWie: Channel-wise rgb-mean subtraction for each input (note NCHW --> NTCHW reduce over NCHW).

examples/OpticalFlow/flownet2s.py
PatWie: Yep, not for now.

PatWie:
The weights of FlowNet-S are use. FlowNet-s is a thin version.

> For the stack configuration we append upper- or lower-case letters to indicate the original
FlowNet or the thin version with 3/8 of the channels.

PatWie:
Added FlowNet2-C. Until they merged this [PR](https://github.com/tensorflow/tensorflow/pull/21392) somehow. To keep this C++ stuff reasonable small, I just included a 1:1 refactoring of the pytroch version with giving credits.

PatWie:
I think everything is complete for now. The training protocol seems to be just tedious to re-implement. But this would be another PR.

ppwwyyxx:
I hope we don't include any ops. I agree with https://github.com/tensorflow/tensorflow/pull/21392#issuecomment-412354516 that it's not easy to maintain external ops and let it build automatically. After copying the missing CMakeLists.txt from your repo it still takes me about 10 minutes to get things running. I have to deal with:
1. Set the compiler that nvcc uses to gcc <= 5
2. TF version must match exactly. Did some hacks but seems unable to make it work with my TF 1.9.0rc1. Reinstall TF 1.10
3. download cub and set path
4. change ``--gpu-architecture``.

There shouldn't be so many manual steps to get an example running.

ppwwyyxx:
So they call it a "patch" but only use `kernel_size=1`?
In this case it's not too bad to write it in python with a for loop over D^2 slices, compute each slice and concat them in the end. The op is only used once so the speed may not be a huge concern.

PatWie:
Wow, what an incredible smart move on my side. I didn't saw the CMakeLists.txt being in the .gitignore file :joy:

I started a rudimentary python implementation but gave it up in favor of the CPU version. 

1. That's only required when using CUDA >=9.1 (in an incompatible combination with Ubuntu18.04)
2. TF 1.9, 1.10 stable versions work (was a TensorFlow issue before)
3. is now written in the Readme.md
4. I don't see this point. It works fine even on a GTX 960.

Another workaround would be just the CPU version, which reduces all your points to "2". If I find some free time, I could hack a fallback option as pure TensorFlow.

ppwwyyxx:
But a CPU version would be slower than a python version running on a GPU. I made an implementation although I'm not very sure about its correctness. It's about 3~5x slower on GPU
```python
from __init__ import correlation_cost

h = 64
w = 64
CHAN = 256
BATCH = 4

shape = [BATCH, CHAN, h, w]
ina = tf.random_normal(shape, dtype=tf.float32)
inb = tf.random_normal(shape, dtype=tf.float32)

d = 20
D = 21
assert d % 2 == 0

out = correlation_cost(
        ina, inb, 1, d, stride_1=1, stride_2=2, pad=d,
        data_format='NCHW')

inb = tf.pad(inb, [[0, 0], [0, 0], [d, d], [d, d]])

res = []
for k1 in range(0, D):
    start_h = k1 * 2
    for k2 in range(0, D):
        start_w = k2 * 2
        s = tf.slice(inb, [0, 0, start_h, start_w], [-1, -1, h, w])
        ans = tf.reduce_mean(ina * s, axis=1, keepdims=True)
        res.append(ans)
res = tf.concat(res, axis=1)   # ND^2HW


def bench(op, warmup, iter):
    for k in range(warmup):
        op.run()
    start = time.perf_counter()
    for k in range(iter):
        op.run()
    duration = time.perf_counter() - start
    return duration

sess = tf.Session()
with sess.as_default():
    o1, o2 = sess.run([out, res])
    print(np.abs(o1 - o2).max())

    print(bench(out.op, 10, 20))
    print(bench(res.op, 10, 20))
```

PatWie:
Not sure, what you are doing, but the CPU version has 8 nested for-loops. A working TensorFlow implemenation is:

```python
def correlation(A, B, kernel_size, max_displacement, stride_1, stride_2, pad, data_format):
    import numpy as np
    """ This is a fallback option for the correlation cost layer
    """
    assert kernel_size == 1
    assert data_format == 'NCHW'

    b, c, h, w = A.shape.as_list()

    r = max_displacement / stride_2
    d = 2 * r + 1
    border = max_displacement
    dr = int(max_displacement / stride_2)

    Cout = int(d * d)
    Hout = int(np.ceil((h + 2 * (pad - border)) / stride_1))
    Wout = int(np.ceil((w + 2 * (pad - border)) / stride_1))

    Apad = tf.pad(A, [[0, 0], [0, 0], [pad, pad], [pad, pad]])
    Bpad = tf.pad(B, [[0, 0], [0, 0], [pad, pad], [pad, pad]])

    res = []

    for tj in range(-dr, dr + 1):
        for ti in range(-dr, dr + 1):
            res_h = []
            for h in range(0, Hout):
                h1 = int(h * stride_1 + max_displacement)
                res_w = []
                for w in range(0, Wout):
                    w1 = int(w * stride_1 + max_displacement)

                    patchA = Apad[:, :, h1:h1+1, w1:w1+1]

                    w2 = w1 + ti * stride_2
                    h2 = h1 + tj * stride_2

                    patchB = Bpad[:, :, h2:h2+1, w2:w2+1]

                    ans = tf.reduce_mean(patchA * patchB, axis=[1], keepdims=True)
                    res_w.append(ans)

                res_h.append(tf.concat(res_w, axis=3))
            res.append(tf.concat(res_h, axis=2))
    res = tf.concat(res, axis=1)
    return res
```

But this requires H and W dimension to be known beforehand (not `None`)

ppwwyyxx:
What you did look similar to my code. Mine was specialized for stride_1=1 and stride_2=2.

PatWie:
Your code is fine! :+1: 

I guess we need another version for [PWC](https://github.com/PatWie/tensorflow-recipes/blob/master/OpticalFlow/pwc.py) then.

examples/OpticalFlow/flownet_models.py
```diff
@@ -0,0 +1,420 @@
...
+        prediction = self.graph_structure(x)
+        tf.identity(prediction, name="prediction")
+
+    def graph_structure(self, x):
+        x1, x2 = tf.unstack(x, axis=2)
+        x1x2 = tf.concat([x1, x2], axis=1)
+
+        # FlowNet-C
+        flownetc_flow2 = FlowNet2C().graph_structure(x)
+        flownetc_flow = resize(flownetc_flow2 * 20., mode='bilinear')
```
ppwwyyxx: Could you make "20" a constant variable, and add some words and references about it? That would make me more comfortable seeing it around.
Also, is it redundant here to multiple it by 20 and then divide by 20 later?

examples/OpticalFlow/flownet_models.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def inputs(self):
+        return [tf.placeholder(tf.float32, (1, self.channels, self.height, self.width), 'left'),
+                tf.placeholder(tf.float32, (1, self.channels, self.height, self.width), 'right')]
+
+    def graph_structure(self, x):
+        raise NotImplementedError()
+
+    def build_graph(self, left, right):
+        x = tf.stack([left, right], axis=2)
```
ppwwyyxx: It seems to me that all the other parts of the model uses the [2, C, H, W] layout. Why using a different layout here?

You can use [2, C, H, W] here to compute rgb_mean, and let `graph_structure` take [2, C, H, W]. I think this will remove many of the concat+stack

PatWie:
I add the note about the mysterious 20 (which is essentially the max_displacement the network can handle). I not gonna touch any operation or change them. The version works fine. Feel free to change anything.

AEE on Sintel-Clean of *2.10* (authors report 2.03). 
Despite all effort, there must be a different between Caffe and TensorFlow. But I guess this difference is ok. For the evaluation I followed [this implementation](https://github.com/NVIDIA/flownet2-pytorch/blob/6827b4c59d1b5d58bcce492e2d5ec2d6cd1e42de/datasets.py#L68-L70) and simply crop the input images.

PatWie:
I hope I cherry picked all commits during rebasing on master.

examples/OpticalFlow/flownet_models.py
PatWie: Same response as above. I did a step-by-step debugging/comparison against Caffe. Changing these lines would destroy this validation script. But feel free to change it. 

My delivered version yields comparable AEE (with the tolerance one can expect between frameworks). If the AEE stays the same: I am ok.

ppwwyyxx:
Thanks! I can reproduce the 2.1 results. But with flownet2-s:
```
python3 flownet2.py --load flownet2-s.npz  --model flownet2-s --sintel_path ~/data/Sintel/training/
```
I got an error of 13.365433.

And with flownet2-c, I saw:
```
WRN The following variables are in the graph, but not found in the dict: upsampled_flow3_to_2/bias:0, upsampled_flow4_to_3/bias:0, upsampled_flow5_to_4/bias:0, upsampled_flow6_to_5/bias:0
```

PatWie:
Yes, for FlowNet-S there are some waaaaaaay out of range. However, on a few testing examples against the reference implementation I get:

```
actual (1, 2, 96, 128)  expected  (1, 2, 96, 128)
sum-error  0.0009913198
mean-error  3.323065428514557e-14
```

I will test those

flow0264
flow0314
flow0248
flow0002
flow0061

This is a little bit intriguing as all these modules are used in FlowNet2 itself. But you are right according the paper the numbers should be somewhere around ~4.50.

PatWie:
Weights are now fixed (the missing weights for flownet-c) and the model is fixed :unamused:
Sometimes a single character makes a difference.

I uploaded a new version of (flownet-c.npz). I tested (downloaded the weights) and started the runs for FlowNet-S and FlowNet-C (results are below):

| Model|  AEE(sintel clean) |
| ------ | ------ |
| FlowNet-S | 3.82|
| FlowNet-C | 3.08|
| FlowNet2 | 2.10 (authors report 2.03) |

PR #872, opened 2018-08-26
make dataflow idiomatic python container objects (fix #869)

### LINKED ISSUE #869: [proposal] DataFlow design
I propose to change the design of the dataflow to be more concise when using the python idiomatic variant of:

```python
class MyFlow(DataFlow):
   def __init__(self, ...):
   def __iter__(self, ...):
   def __len__(self, ....): 
```

This would make the code somewhat more consistent to other python objects:

```python
class SomeData(object):
  def __init__(self, start, len=10):
    super(SomeData, self).__init__()
    self.start = start
    self.len = len

  def __len__(self):
    return self.len

  def __iter__(self):
    for i in range(self.start, self.start + self.len):
      yield [i]


class AddOne(object):
  """docstring for AddOne"""

  def __init__(self, other_ds):
    super(AddOne, self).__init__()
    self.other_ds = other_ds

  def __len__(self):
    return len(self.other_ds)

  def __iter__(self):
    for i in self.other_ds:
      yield [i[0] + 1]


ds = SomeData(2, 3)
ds = AddOne(ds)
for dp in ds:
  print(dp)
```

In fact, all `dataflow.*`  objects are just glorified [container types](https://docs.python.org/3/reference/datamodel.html#emulating-container-types) (ok most of them). So why not using the python terminology then? To keep backward compatibility, the `get_data()` and `size()` can wrap the `__iter__` and `__len__`. I would advise against using `__next__` to not bloating the code.

I could do a pass and apply the changes accordingly, but I am not eager to open yet another PR -- and even more so when not discussed beforehand.

Maybe even `__str__` can take over the implementation of `PrintData`. I see currently no usage of PrintData outside debugging.

```python
ds = SomeData(2, 3)
print(ds)
```

seems a better fit and more descriptive than

```python
ds = SomeData(2, 3)
ds = PrintData(ds)
next(ds.get_data())
```

The only problem I can see is when using `PrintData` at multiple places. But again, `PrintData` shouldn't be deprecated. Instead `__str__` can be an additional feature. Or even better, `print(ds)` can print the entire dataflow information of all stacked dataflow mappings.

PatWie:
Change `DataFlow` to support `__iter__, __len__`.


tensorpack/dataflow/base.py
```diff
@@ -39,29 +40,59 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+    def __init__(cls, name, bases, dct):
+        super(DataFlowMeta, cls).__init__(name, bases, dct)
+
+
+@six.add_metaclass(DataFlowMeta)
 class DataFlow(object):
     """ Base class for all DataFlow """
 
     @abstractmethod
```
ppwwyyxx: When the metaclass is not ABCMeta I guess `abstractmethod` will have no effect.
Instead we should check in our own meta, that one of the two methods `get_data` and `__iter__` was implemented.

tensorpack/dataflow/base.py
```diff
@@ -39,29 +40,59 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    """
+    DataFlow uses "__iter__()" and "__len__()" instead of
+    "get_data()" and "size()". This add back-compatibility.
+    """
+    def __new__(meta, name, bases, dct):
+        # "'get_data' in dct" is important as we just care about DataFlow itself
+        # and not derived classes.
+        if '__iter__' not in dct and 'get_data' in dct:
+            dct['__iter__'] = dct['get_data']
+            log_deprecated("DataFlow.get_data()",
```
ppwwyyxx: Let's deprecate them after a while. For big changes like this let's first apply it and see what happened.

tensorpack/dataflow/base.py
```diff
@@ -39,29 +40,59 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                           "use DataFlow.__iter__() instead!",
+                           "2018-12-30")
+        if '__len__' not in dct and 'size' in dct:
+            dct['__len__'] = dct['size']
+            log_deprecated("DataFlow.size()",
+                           "use DataFlow.__len__() instead!",
+                           "2018-12-30")
+        return super(DataFlowMeta, meta).__new__(meta, name, bases, dct)
+
+    def __init__(cls, name, bases, dct):
```
ppwwyyxx: This is probably not necessary though I haven't tested

tensorpack/callbacks/inference_runner.py
```diff
@@ -61,15 +61,15 @@ class InferenceRunnerBase(Callback):
     """ Base class for inference runner.
 
     Note:
-        1. InferenceRunner will use `input.size()` to determine
+        1. InferenceRunner will use `input.__len__()` to determine
```
ppwwyyxx: Why not `len(x)` but `x.__len__()` ?

tensorpack/utils/utils.py
```diff
@@ -214,7 +214,9 @@ def get_tqdm_kwargs(**kwargs):
     return default
 
 
-def get_tqdm(**kwargs):
+def get_tqdm(df=None, **kwargs):
```
ppwwyyxx: This is not needed. `get_tqdm(iterable=df)` is the same.

ppwwyyxx: Or maybe add `*args, **kwargs`.

tensorpack/dataflow/base.py
PatWie: My terminal is flashing anyway because of `python-prctl`. A few more warning would be ok.

tensorpack/dataflow/base.py
PatWie: I just use `ABCMeta` as the parent and `@abstractmethod` is working

tensorpack/dataflow/base.py
PatWie: confirm and removed

tensorpack/callbacks/inference_runner.py
PatWie: fixed.

tensorpack/utils/utils.py
PatWie: `*args, **kwargs` looks good!

tensorpack/dataflow/serialize.py
```diff
@@ -99,11 +99,9 @@ def save(df, path):
             path (str): output npz file.
         """
         buffer = []
-        size = _reset_df_and_get_size(df)
-        with get_tqdm(total=size) as pbar:
-            for dp in df.get_data():
-                buffer.append(dp)
-                pbar.update()
+        df.reset_state()
+        for dp in get_tqdm(df):
```
ppwwyyxx: This will fail if df has no size. `_reset_df_and_get_size(df)` was meant to provide a default size of 0.

ppwwyyxx:
I revert the change of interface on InputSource. I think __len__ doesn't apply very well since InputSource is not really iterable. The size is more like an attribute of it.

Also, from https://docs.python.org/3/library/functions.html#len:


> len(s)
Return the length (the number of items) of an object. The argument may be a sequence (such as a string, bytes, tuple, list, or range) or a collection (such as a dictionary, set, or frozen set).

PR #885, opened 2018-09-04
Example: Exporting models from Tensorpack

PatWie:
The old `export` function had no documentation. Further, the entire process can benefit from using the `PredictConfig`.

- [x] refactor TensorFlow-Serving exporter to use `PredictConfig`
- [x] small and meaningful example with a difference between train and inference graph
- [x] add export to pruned/frozen graph
- [x] add documentation
- [x] example which demonstrates, export, debugging, loading of several export options.


ppwwyyxx:
Honestly I don't like the tf-lite and mobile thing in general because my impression is that they are premature and won't work well for any actual model I care about..

PatWie:
I do not really refer to tf-lite or tf-mobile. I refer to a frozen graph (all constant) and pruned graph (only nodes necessary for inference).
In all my cases this worked perfectly and I can restore even graphs which I trained several months ago without any issues.

So far, the documentation is missing but the minimal example shows probably the most important things.

ppwwyyxx:
Last time I tried those graph transform, it does not do what I want (fuse BN into conv layers) and I end up rewriting the graph and transformed the weights manually.

ppwwyyxx:
The features to be added to the library by this PR seems to be just: exporting the model to saved_model format (for serving) or a pruned graph format.
I think the features make sense as part of a training framework because the output can be seen as a result of the training. After exporting to standard format a user can then do other fancy stuff such as toco or tensorrt conversion which I don't want to include in tensorpack.
For implementation I think it's simpler to just have the old name (`ModelExporter`) and two methods for two different formats. Also the pruned graph does not seem to be very related to "mobile" so maybe a better name is needed?

PatWie:
Merge both classes into `ModelExport` and renamed *mobile* to *compact* to represent a compressed and self-contained graph. The example is updated as well to reflect those changes.

tensorpack/tfutils/export.py
```diff
@@ -1,71 +1,85 @@
...
+            input_tensors = get_tensors_by_names(self.config.input_names)
+            output_tensors = get_tensors_by_names(self.config.output_names)
+
+            self.config.session_init._setup_graph()
+            # we cannot use "self.config.session_creator.create_session()" here since it finalizes the graph
+            sess = tf.Session(config=tf.ConfigProto(allow_soft_placement=True))
+            self.config.session_init._run_init(sess)
+
+            if dtypes is None:
+                dtypes = [n.dtype for n in input_tensors]
```
ppwwyyxx: Is there a reasonable use case for this `dtypes` option? (i.e. to not use the default?)

tensorpack/tfutils/export.py
```diff
@@ -1,71 +1,85 @@
...
-        self.output_names = output_names
-        self.input_names = input_names
+    def export_compact(self, filename, dtypes=None):
+        """Create a self-contained inference-only graph and write final graph to disk.
 
-    def export(self, checkpoint, export_path,
-               tags=[tf.saved_model.tag_constants.SERVING],
-               signature_name='prediction_pipeline'):
+        Args:
+            export_graph_file (str): path to final local of the graph
```
ppwwyyxx: please fix this comment

tensorpack/tfutils/export.py
```diff
@@ -1,71 +1,85 @@
...
+                False)
+
+            with gfile.FastGFile(filename, "wb") as f:
+                f.write(pruned_graph_def.SerializeToString())
+
+    def export_serving(self, filename,
+                       tags=[tf.saved_model.tag_constants.SERVING],
+                       signature_name='prediction_pipeline'):
+        """
+        Converts and checkpoint to a servable for TensorFlow Serving.
```
ppwwyyxx: Do you mean "converts graph and checkpoint"?

tensorpack/tfutils/export.py
PatWie: For images `InferenceModel` the `dtype` here needs to be `tf.uint8` although the placeholder has type `tf.string`. There are some strange corner cases, which I had.

ppwwyyxx: What is the issue? If the dtype parameter for `optimize_for_inference_lib.optimize_for_inference` cannot match `tensor.dtype`, that sounds like a tensorflow bug to me, is it?

examples/basics/export.py
```diff
@@ -0,0 +1,260 @@
...
+        # outputs should be base64 encoded strings agains
+        prediction_img = tf.clip_by_value(prediction_img, 0, 255)
+        prediction_img = tf.cast(prediction_img, tf.uint8)
+        prediction_img_bytes = tf.map_fn(tf.image.encode_png, prediction_img, dtype=tf.string)
+
+        # prediction_img_bytes = tf.image.encode_png(prediction_img)
+        tf.identity(prediction_img_bytes, name='prediction_img_bytes')
+
+
+def export_npz():
```
ppwwyyxx: I would like to remove this function. Npz is used for the model zoo because it's easier to load and only one file. But I don't mean to make it feel like "standard for tensorpack" and users may use any format they like given the standard TF checkpoint support.
And this function isn't doing anything useful, after all.

tensorpack/tfutils/export.py
PatWie: I cannot reproduce it as:

```python
import tensorflow as tf
from tensorflow.python.tools import optimize_for_inference_lib
from tensorflow.python.framework import graph_util
from tensorflow.python.platform import gfile

p = tf.placeholder(tf.string, (None), name='input_img_bytes')
x = tf.image.decode_png(p, channels=3)
x = tf.expand_dims(x, axis=0)

unused = tf.cast(x, tf.float32)
unused = tf.layers.conv2d(unused, 3, name='conv', kernel_size=[1, 1])

x = tf.squeeze(x, axis=0)
x = tf.image.encode_png(x)
o = tf.identity(x, name='final_bytes')

with tf.Session() as sess:
  sess.run(tf.global_variables_initializer())
  graph_def = sess.graph.as_graph_def()

  graph_def = graph_util.convert_variables_to_constants(
      sess, graph_def, ['final_bytes'])

  the_dtype = p.dtype.as_datatype_enum
  graph_def = optimize_for_inference_lib.optimize_for_inference(graph_def,
                                                                ['input_img_bytes'], ['final_bytes'], [the_dtype], False)

  with gfile.FastGFile('/tmp/tmp_graph.pb', "wb") as f:
    f.write(graph_def.SerializeToString())

tf.reset_default_graph()

# ---------inference ------------------------

with tf.Session(config=tf.ConfigProto(allow_soft_placement=True)) as sess:
  with tf.gfile.GFile('/tmp/tmp_graph.pb', "rb") as f:
    graph_def = tf.GraphDef()
    graph_def.ParseFromString(f.read())
    tf.import_graph_def(graph_def)

  print([n.name for n in sess.graph.as_graph_def().node])

  input_img = sess.graph.get_tensor_by_name('import/input_img_bytes:0')
  prediction_img = sess.graph.get_tensor_by_name('import/final_bytes:0')

  with open('lena.png', 'rb') as f:
    buf = bytes(f.read())

  prediction = sess.run(prediction_img, {input_img: buf})

  with open('lena2.png', 'wb') as f:
    f.write(prediction)
```

works fairly well. However, I had some troubles. I am fine with removing the argument. The issues is probably related to `tf.map_fn`.

examples/basics/export.py
PatWie: This was just a mental note for myself.

tensorpack/tfutils/export.py
realwill: I got error when froze mask rcnn with optimize_for_inference_lib.optimize_for_inference:

File "inference.py", line 35, in <module>
    tf.import_graph_def(od_graph_def)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/util/deprecation.py", line 488, in new_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/tensorflow/python/framework/importer.py", line 422, in import_graph_def
    raise ValueError(str(e))
ValueError: NodeDef expected inputs '' do not match 1 inputs specified; Op<name=Const; signature= -> output:dtype; attr=value:tensor; attr=dtype:type>; NodeDef: {{node import/output/map/while/add/y}} = Const[dtype=DT_INT32, value=Tensor<type: int32 shape: [] values: 1>](import/output/map/while/Switch:1)

ppwwyyxx: For anyone to better understand and diagnose your issue, please post relevant details following the issue template (click "New Issue" -> "Unexpected Problems / Bugs", or visit [this link to post issues about unexpected problems](https://github.com/tensorpack/tensorpack/issues/new?template=unexpected-problems---bugs.md))

PR #917, opened 2018-10-04
simplify GAN usage

PatWie:
I always struggle to remember the order of arguments in 

```python
if nr_tower == 1:
  trainer = GANTrainer(data, Model())
else:
  trainer = MultiGPUGANTrainer(nr_tower, data, Model())
```

### DIFF
--- a/examples/GAN/GAN.py
+++ b/examples/GAN/GAN.py
@@ -10,6 +10,7 @@ from tensorpack.tfutils.tower import TowerContext, TowerFuncWrapper
 from tensorpack.graph_builder import DataParallelBuilder, LeastLoadedDeviceSetter
 from tensorpack.tfutils.summary import add_moving_summary
 from tensorpack.utils.argtools import memoized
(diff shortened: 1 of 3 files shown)


examples/GAN/GAN.py
```diff
@@ -72,40 +73,89 @@ def get_optimizer(self):
...
         # Setup input
-        cbs = input.setup(inputs_desc)
+        cbs = input.setup(model.get_inputs_desc())
         self.register_callback(cbs)
 
-        """
-        We need to set tower_func because it's a TowerTrainer,
-        and only TowerTrainer supports automatic graph creation for inference during training.
+        if num_gpu < 2:
+            _build_single_gan_trainer(self, input, model)
```
ppwwyyxx: Passing `self` around is an unusual design. Here we can just write them as methods of `GANTrainer`.

PatWie: You are correct. I also made a mistake with the StagingInput, which is fixed now.

PR #924, opened 2018-10-10
Two updates

ksellesk:
1. Add BNLeakyReLU

2. Fix valid_for_nccl: nccl supports fp16 now


ppwwyyxx:
Please remove BNLeakyReLU. Tensorpack will not add any symbolic layers. The existing layers exists only because there were no alternatives at the beginning. See more at the tutorial: https://tensorpack.readthedocs.io/tutorial/symbolic.html

PR #926, opened 2018-10-10
Fix valid_for_nccl: nccl supports fp16 now

ksellesk:


ppwwyyxx:
It seems it's supported since TF1.8: https://github.com/tensorflow/tensorflow/commit/5c469e6bafb479ef110b2f02f070507a3711664d#diff-280d77a571d7e6e2e5c11563762014b2
Please `from tensorpack.tfutils.common import get_tf_version_tuple` and only enable this feature when the version matches.

PR #928, opened 2018-10-11
Use last k observations instead of last k epochs.

thuzhf:


tensorpack/callbacks/param.py
```diff
@@ -373,10 +371,11 @@ def _get_value_to_set(self):
             return None
 
         self.history.append(last)
+        self.observations_since_last_changed += 1
 
         if len(self.history) < self.history.maxlen or \
-                self.epoch_num - self.last_changed_epoch < self.history.maxlen:
-            # not full yet, or value have changed just now
+                self.observations_since_last_changed < self.history.maxlen:
+            # not full yet, or the k observations which lead to the last change haven't been all popped out
```
ppwwyyxx: Is it better if we just clear the history?

tensorpack/callbacks/param.py
```diff
@@ -151,12 +151,10 @@ def get_value_to_set(self):
         """
         ret = self._get_value_to_set()
         if ret is not None and ret != self._last_value:
-            if self.epoch_num != self._last_epoch_set:
-                # Print this message at most once every epoch
-                logger.info("[HyperParamSetter] At global_step={}, {} will change to {:.8f}".format(
-                    self.global_step, self.param.readable_name, ret))
+            logger.info("[HyperParamSetter] At global_step={}, {} will change to {:.8f}".format(
```
ppwwyyxx: I tried to avoid printing this information too often because of how noisy it can be. But I understand there is a confusion when it's only printed once per epoch. Maybe the message or the docs can be improved.

tensorpack/callbacks/param.py
thuzhf: Yes, it seems more concise.

tensorpack/callbacks/param.py
thuzhf: OK, I understand..

thuzhf: Do you want me to keep this change or revert this change in order for you future change on this part?

PR #934, opened 2018-10-18
fix dp extraction bug for visualize().

cykustcc:
dp[0] will raise KeyError.

### DIFF
--- a/examples/FasterRCNN/train.py
+++ b/examples/FasterRCNN/train.py
@@ -349,11 +349,11 @@ def visualize(model, model_path, nr_visualize=100, output_dir='output'):
     utils.fs.mkdir_p(output_dir)
     with tqdm.tqdm(total=nr_visualize) as pbar:
         for idx, dp in itertools.islice(enumerate(df), nr_visualize):
-            img = dp[0]
+            img = dp['image']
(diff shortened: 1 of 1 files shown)


examples/FasterRCNN/train.py
```diff
@@ -349,11 +349,11 @@ def visualize(model, model_path, nr_visualize=100, output_dir='output'):
...
     with tqdm.tqdm(total=nr_visualize) as pbar:
         for idx, dp in itertools.islice(enumerate(df), nr_visualize):
-            img = dp[0]
+            img = dp['image']
             if cfg.MODE_MASK:
-                gt_boxes, gt_labels, gt_masks = dp[-3:]
+                gt_boxes, gt_labels, gt_masks = dp['gt_boxes'], dp['gt_labels'], dp['gt_masks']
             else:
-                gt_boxes, gt_labels = dp[-2:]
+                gt_boxes, gt_labels = dp['gt_boxes'], dp['gt_labels']
```
ppwwyyxx: Only keeping this one line is enough. The other branch can be removed since `gt_masks` is not used later.

cykustcc: Have fixed this as suggested.

PR #979, opened 2018-11-08
SR sample can take input folder

notecola:
Change SR sample to take folder as input and process all images.

### DIFF
--- a/examples/SuperResolution/enet-pat.py
+++ b/examples/SuperResolution/enet-pat.py
@@ -219,9 +219,42 @@ class Model(GANModelDesc):
         return opt
 
 
+def apply_all(model_path, lowres_path=".", output_path='.'):
+    assert os.path.isdir(lowres_path)
+    assert os.path.isdir(output_path)
+
+    for filename in os.listdir(lowres_path):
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Your code has a large amount of duplication with the existing code, and is very inefficient because it builds a predictor for each image. Therefore it cannot be accepted.

ppwwyyxx:
Do you still wish to add this functionality?

It's easier to just modify on top of the existing `apply` function. Make it take a list of `lowres` instead of a single one, and evaluate them one by one. Python's argparse can take a list of arguments. This way the usage of the script for single / multiple images will also be compatible.

notecola:
Hi @ppwwyyxx thanks for your response. But what about the Model(...) constructed every time for each image with the image original resolution? Is there any way to have it initialized once?

notecola:
This would work if all input images have the same size. In other words, LR_SIZE_H and LR_SIZE_W are different for each input image.

ppwwyyxx:
Then you'll need to modify the model to support arbitrary size (i.e. None size)

PR #983, opened 2018-11-16
Add support for string batching

yg320:
Hi,

I wasn't able to batch data points of type string (got AttributeError exception). This change fixes that issue.

### DIFF
--- a/tensorpack/dataflow/common.py
+++ b/tensorpack/dataflow/common.py
@@ -135,6 +135,8 @@ class BatchData(ProxyDataFlow):
                     tp = 'int32'
                 elif type(dt) == float:
                     tp = 'float32'
+                elif type(dt) == str:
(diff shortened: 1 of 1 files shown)


tensorpack/dataflow/common.py
```diff
@@ -135,6 +135,8 @@ def _aggregate_batch(data_holder, use_list=False):
                     tp = 'int32'
                 elif type(dt) == float:
                     tp = 'float32'
+                elif type(dt) == str:
```
ppwwyyxx: This behaves differently for Python 2 and 3.
Are you using Python 2 or 3?

ppwwyyxx:
```suggestion
                elif isinstance(dt, (six.binary_type, six.text_type)):
```

PR #984, opened 2018-11-16
update crowd label handeling

bluerythem:
The previous way to thresh anchor boxes based on iou is not a good metric.  For example, small anchors inside large crowd box would have small iou so not being ignored.

This PR update crowd label handling logic. Here, we use ioa instead of iou so that anchors that overlap with crowd boxes could be properly filtered out.

### DIFF
--- a/examples/FasterRCNN/config.py
+++ b/examples/FasterRCNN/config.py
@@ -139,7 +139,10 @@ _C.RPN.FG_RATIO = 0.5  # fg ratio among selected RPN anchors
 _C.RPN.BATCH_PER_IM = 256  # total (across FPN levels) number of anchors that are marked valid
 _C.RPN.MIN_SIZE = 0
 _C.RPN.PROPOSAL_NMS_THRESH = 0.7
-_C.RPN.CROWD_OVERLAP_THRESH = 0.7  # boxes overlapping crowd will be ignored.
+# Anchors which overlap with a crowd box (IOA larger than threshold) will be ignored.
+# Setting this to a value larger than 1.0 will disable the feature.
+# It is disabled by default because Detectron does not do this.
+_C.RPN.CROWD_OVERLAP_THRESH = 9.99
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
As said in the code, we did not do this filtering because the official implementation (Detectron) did not do this.
And in practice I've found no differences.
Did you find any accuracy difference by using your filtering?

> For example, small anchors inside large crowd box would have small iou so not being ignored.

It's arguable whether small anchors inside large crowd box should be ignored.

bluerythem:
I think it is wrong not to handle crowd label even Detectron did not. The network would be penalized for doing true detection inside the crowd label region.
 
I have not tried to train on Coco but for my private dataset which includes about 3% crowd label, it improve about 2 points on overall mAP.

If it did not make a difference for your former training, is it possible that former code based on iou did not properly ignore anchors in crowd box region?
Another possible theory is the crowd region would be marked as bgs if not taken cared of. But later the code samples fgs and bgs and there are many bg labels so some mistakes in crowd label region is may be statistically not significant. 

I am curious why ignoring small anchors inside large crowd box is arguable? Isn't true positives inside crowd box are usually small? Like a person in a crowd of people?

ppwwyyxx:
>  Isn't true positives inside crowd box are usually small? Like a person in a crowd of people?

Crowd is labeled as boxes. In coco the labeled boxes are sometimes unnecessarily big due to lazy labelers. The crowd box may as as result overlap meaningful objects, especially objects of a different category (e.g. a car in a crowd of people).

What you said about crowd boxes all make sense -- that's why there is crowd box handling code, though disabled. And they are disabled because there is importance in maintaining consistency with official code.

Therefore how about let's add your improvement but at the same time still keep things consistent by default.
Could you change the default `CROWD_OVERLAP_THRESH` to something like 1.0 or larger (since it's not used at all for now) and add a comment that this is by default disabled?

ppwwyyxx:
Just made the above changes to this PR. Let me know if you found anything is incorrect

PR #1017, opened 2018-12-18
3D convolution wrapper

andreanicastro:
An attempt to add to tensorpack 3D convolutions. I don't know the entire codebase so it is possible that I might miss quite a lot of stuff. 
This also showed some design issues. See for example argtools.py

### DIFF
--- a/tensorpack/models/conv3d.py
+++ b/tensorpack/models/conv3d.py
@@ -1,0 +1,177 @@
+# -*- coding: utf-8 -*-
+# File: conv3d.py
+
+import tensorflow as tf
+from .common import layer_register, VariableHolder
+from ..tfutils.common import get_tf_version_tuple
+from ..utils.argtools import shape3d, shape5d, get_data_format
+from .tflayer import rename_get_variable, convert_to_tflayer_args
+
+
+__all__ = ['Conv3D', 'Deconv3D', 'Conv3DTranspose']
+
+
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
Please see the following from https://tensorpack.readthedocs.io/tutorial/symbolic.html:

> These layers were written only because there were no alternatives when tensorpack was first developed. Nowadays, these implementation actually call tf.layers directly. Tensorpack will not add any more layers into its core library because this is not the focus of tensorpack, and there are many other alternative symbolic libraries today.

You can use `tf.layers.Conv3D` directly in tensorpack. You do not need to add a tensorpack wrapper for them.

andreanicastro:
I know. That is exactly what I had. But then I wrote the wrapper for the nice logging and argscope. I thought might have been useful. Probably the split implementation is a bit useless though.


On 18 Dec 2018 18:08, Yuxin Wu <notifications@github.com> wrote:

Please see the following from https://tensorpack.readthedocs.io/tutorial/symbolic.html:

These layers were written only because there were no alternatives when tensorpack was first developed. Nowadays, these implementation actually call tf.layers directly. Tensorpack will not add any more layers into its core library because this is not the focus of tensorpack, and there are many other alternative symbolic libraries today.

You can use tf.layers.Conv3D directly in tensorpack. You do not need to add a tensorpack wrapper for them.

Issue #283, opened 2017-06-02
Support of Conv3D and Deconv3D


tmquan:
@Hi tensorpack, 

Is there any plan to support wrapper of Conv3D and Deconv3D (Conv3D_Transpose)?
Currently, I have 2 versions: a self-implemented custom code and tflearn wrapper. But I believe having supports from tensorpack library is much more useful. 

3D version has been recently added to r1.2.0 https://github.com/tensorflow/tensorflow/blob/r1.2/RELEASE.md 

It is somehow related to this issue #211 for people who are working on 3D CNN and 3D FCN

Best,

ppwwyyxx:
If you have such code already, you can call your code or tflearn wrapper directly within a tensorpack model.

ppwwyyxx:
Closing this and tracking the issue in #291.

PR #1048, opened 2019-01-15
Add 3D Conv and supporting pooling and batch_norm for 5D tensor

YongyiTang92:
Add 3D Conv and supporting pooling and batch_norm for 5D tensor related to #283


ppwwyyxx:
It is mentioned in the pull request template https://github.com/tensorpack/tensorpack/blob/master/.github/pull_request_template.md, which you can see when opening a pull request, that we do not add layers into tensorpack.

PR #1049, opened 2019-01-15
argscope_for_function had some issues when a layer produces multiple …

PatWie:
`argscope_for_function` failed for layers with multiple outputs.

### DIFF
--- a/tensorpack/tfutils/argscope.py
+++ b/tensorpack/tfutils/argscope.py
@@ -79,13 +79,22 @@ def enable_argscope_for_function(func, log_shape=True):
             myfunc = enable_argscope_for_function(myfunc)
 
     Args:
-        func: function which should be decorated.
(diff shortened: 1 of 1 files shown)


tensorpack/tfutils/argscope.py
```diff
@@ -97,9 +97,14 @@ def wrapped_func(*args, **kwargs):
         name = func.__name__ if 'name' not in kwargs else kwargs['name']
         if log_shape:
             if ('tower' not in ctx.ns_name.lower()) or ctx.is_main_training_tower:
+                # we assume the first parameter is the most intersting
```
ppwwyyxx: If it's really going to handle arbitrary user function, it's best to (1) document what types of functions is supported in the documentation and (2) add a check for it (either skip the logging or raise an exception, both seems OK)

PR #1050, opened 2019-01-15
Add the compression parameter to Horovod trainer

r96725046:


tensorpack/train/trainers.py
```diff
@@ -6,6 +6,7 @@
 import sys
 import tensorflow as tf
 
+from horovod.tensorflow.compression import Compression
```
ppwwyyxx: Cannot import here because horovod is not a required dependency of tensorpack.

tensorpack/train/trainers.py
```diff
@@ -370,10 +371,11 @@ class HorovodTrainer(SingleCostTrainer):
            for a full example which has handled these common issues.
            This example can train ImageNet in roughly an hour following the paper's setup.
     """
-    def __init__(self, average=True):
+    def __init__(self, average=True, compression=Compression.none):
```
ppwwyyxx: The default value should be `None` because `Compression` may not be imported when horovod is not installed.

tensorpack/train/trainers.py
```diff
@@ -370,10 +371,11 @@ class HorovodTrainer(SingleCostTrainer):
...
         """
         Args:
             average (bool): whether to average or sum the gradients across processes.
+            compession: None, fp16
```
ppwwyyxx: fix typo

tensorpack/train/trainers.py
```diff
@@ -399,7 +402,7 @@ def allreduce(self, grads):
         with tf.name_scope("HVDAllReduce"):
             for grad, var in grads:
                 if grad is not None:
-                    avg_grad = hvd.allreduce(grad, average=self._average)
+                    avg_grad = hvd.allreduce(grad, average=self._average, compression=self._compression)
```
ppwwyyxx: should add a check because this feature is only supported after horovod 0.15

r96725046: Sure, I found it only works after horovod 0.15

tensorpack/train/trainers.py
r96725046: Will put a try catch block.

ppwwyyxx: Please put the import where horovod is currently imported

r96725046: Move into the HorovodTrainer

PR #1058, opened 2019-01-21
Added MUNIT.py example

Skylion007:
I reimplemented the [MUNIT paper](https://arxiv.org/abs/1804.04732) in Tensorpack. 

I've been able to replicate their results on their cat2dog datasets as well as some of my own ones.

Suggestions welcome.

### DIFF
--- a/examples/GAN/MUNIT.py
+++ b/examples/GAN/MUNIT.py
@@ -1,0 +1,460 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# File: MUNIT.py
+# Author: Aaron Gokaslan
+
+import os
+import argparse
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
I apologize that I did not find enough time to learn the paper and this code.
If you're still interested in adding this example, here are some high-level comments:
1. There is a significant amount of code duplication with CycleGAN, and ideally this should be avoided. I don't know yet what's the most beautiful way to do it though -- probably putting the shared part into a separate file and import it.
2. I need the instructions to run it (including getting the datasets). And some nice-looking sample outputs at certain steps (or when it finishes), so I know what to expect when I test the code.

PR #1151, opened 2019-04-16
Prevent install python-prctl outside Linux when installing by wheel

### LINKED ISSUE #1148: python-prctl platform specifier is useless when installing from a wheel
When installing `tensorpack[all]` on Mac `python-prctl` is expected to be skipped using a platform specifier. Unfortunately this works only when installing from a source package, not from a wheel. The reason is that the platform is evaluated in [setup.py](https://github.com/tensorpack/tensorpack/blob/master/setup.py#L60) at the machine which built the wheel, not at the target machine. This affects the `master` at the time of creating this issue.

```
mac$ pip install tensorpack[all]==0.9.4
# [...]
Collecting python-prctl; extra == "all" (from tensorpack[all]==0.9.4)
  Using cached https://files.pythonhosted.org/packages/7a/90/61935e2530a76f41e9e4f8ba0fe073d4ad0a3e16c4953156253f939fb057/python-prctl-1.7.tar.gz
    Complete output from command python setup.py egg_info:
    This module only works on linux
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/k6/pp54bs857851lpcsrb51fm1h0000gn/T/pip-install-fda0nlll/python-prctl/
```

This works:
```
mac$ pip install tensorpack[all]==0.9.4 --no-binary tensorpack
```

I'm not sure if the proper platform specifiers work in `install_requires`:

```
# this works well in requirements.txt
python-prctl==1.7; 'linux' in sys_platform
```

bzamecnik:
Closes #1148.

When `tensorpack` was installed by wheel the platform was already evaluated
and thus `python-prctl` was installed, even though it is Linux-specific.
Installing from source package or sources was OK.

Instead use a platform specifier that is evaluated while installing the wheel.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,8 +56,8 @@ setup(
     ],
     tests_require=['flake8', 'scikit-image'],
     extras_require={
-        'all': ['pillow', 'scipy', 'h5py', 'lmdb>=0.92', 'matplotlib', 'scikit-learn'] +
-               ['python-prctl'] if platform.system() == 'Linux' else [],
+        'all': ['pillow', 'scipy', 'h5py', 'lmdb>=0.92', 'matplotlib', 'scikit-learn'],
+        'all: "Linux" in sys_platform': ['python-prctl'],
         'all: python_version < "3.0"': ['tornado'],
     },
 )


ppwwyyxx:
Thanks!

PR #1201, opened 2019-05-20
ThroughputTracker

### LINKED ISSUE #1197: Add Throughput Tracking Callback
I would like to add a callback to Tensorpack that can be used to track training throughput, similar to GPUUtilizationTracker or the Host/GPU MemoryTrackers. I think this is a missing feature that is generic enough that it belongs in core Tensorpack.

I would use [this](https://github.com/armandmcqueen/tensorpack-mask-rcnn/blob/master/MaskRCNN/performance.py#L113) as my (rough) starting point. If you agree that this feature belongs in core Tensorpack, I'll open an initial PR for review.

ppwwyyxx:
resolve #1197 
@armandmcqueen does this suffice what you need?
You can make it run more frequent than 1 epoch by `PeriodicTrigger(ThroughputTracker(), every_k_steps=100)`.

### DIFF
--- a/examples/FasterRCNN/train.py
+++ b/examples/FasterRCNN/train.py
@@ -86,6 +86,7 @@ if __name__ == '__main__':
         ScheduledHyperParamSetter('learning_rate', lr_schedule),
         GPUMemoryTracker(),
         HostMemoryTracker(),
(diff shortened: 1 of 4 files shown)


armandmcqueen:
Yes, that looks perfect. Thanks!

PR #1207, opened 2019-05-23
fix "mnist-keras.py"

see--:
Otherwise I see the following error:
```
InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'conv2d_input' with dtype float and shape [?,28,28,1]
	 [[node conv2d_input (defined at examples/keras/mnist-keras.py:45) ]]
```
This way keras does not create an unnecessary Input and determines the shape from [`image`](https://github.com/tensorpack/tensorpack/blob/master/examples/keras/mnist-keras.py#L69)

### DIFF
--- a/examples/keras/mnist-keras.py
+++ b/examples/keras/mnist-keras.py
@@ -42,7 +42,7 @@ def clear_tower0_name_scope():
 def get_keras_model():
     with clear_tower0_name_scope():
         M = keras.models.Sequential()
-        M.add(KL.Conv2D(32, 3, activation='relu', input_shape=[IMAGE_SIZE, IMAGE_SIZE, 1], padding='same'))
+        M.add(KL.Conv2D(32, 3, activation='relu', padding='same'))
         M.add(KL.BatchNormalization())
         M.add(KL.MaxPooling2D())
         M.add(KL.Conv2D(32, 3, activation='relu', padding='same'))


ppwwyyxx:
Which version of tensorflow you test it with?

see--:
'1.13.1' from `pip`

see--:
What do you think about these changes?
```
diff --git a/examples/keras/mnist-keras.py b/examples/keras/mnist-keras.py
index 2dcdd41..95a1af2 100755
--- a/examples/keras/mnist-keras.py
+++ b/examples/keras/mnist-keras.py
@@ -59,14 +59,14 @@ def get_keras_model():
 class Model(ModelDesc):
     def inputs(self):
         return [tf.TensorSpec((None, IMAGE_SIZE, IMAGE_SIZE), tf.float32, 'input'),
+                tf.TensorSpec((), tf.bool, 'learning_phase'),
                 tf.TensorSpec((None,), tf.int32, 'label')]
 
-    def build_graph(self, image, label):
+    def build_graph(self, image, learning_phase, label):
         image = tf.expand_dims(image, 3) * 2 - 1
         ctx = get_current_tower_context()
-
         M = get_keras_model()
-        logits = M(image)
+        logits = M(image, training=learning_phase)
         if ctx.is_main_training_tower:
             for op in M.updates:
                 tf.add_to_collection(tf.GraphKeys.UPDATE_OPS, op)
@@ -97,7 +97,9 @@ class Model(ModelDesc):
 
 def get_data():
     train = BatchData(dataset.Mnist('train'), 128)
+    train = MapData(train, lambda ds: [ds[0], 1, ds[1]])
     test = BatchData(dataset.Mnist('test'), 256, remainder=True)
+    test = MapData(test, lambda ds: [ds[0], 0, ds[1]])
     return train, test
 
 
@@ -109,7 +111,6 @@ if __name__ == '__main__':
         model=Model(),
         dataflow=dataset_train,
         callbacks=[
-            KerasPhaseCallback(True),   # for Keras training
             ModelSaver(),
             InferenceRunner(
                 dataset_test,
```
One could use arbitrary callbacks, `InferenceRunner` is not required.

ppwwyyxx:
I do not see any error running `mnist-keras.py` with TF 1.13.1 on CPU. Could you post details following [the issue template](https://github.com/tensorpack/tensorpack/issues/new?template=unexpected-problems---bugs.md))?

> What do you think about these changes?

What is this for?

ppwwyyxx:
Sorry I was using the wrong version of code. Now I can reproduce the error.

see--:
Thanks!
> What is this for?
I need a custom callback that is not an instance of `InferenceRunner` (like in FasterRCNN example). I think the `KerasPhaseCallback(True)` won't set the model in the inference mode. Using the learning phase as input will.

ppwwyyxx:
You can feed `{keras.backend.learning_phase():False}` in your custom callback when you evaluate a tensor.

ppwwyyxx:
Would you mind sharing why do you want to do these on the `mnist-keras.py` example? The same thing can be done much easier if using tensorpack layers or `tf.layers`, and without any of these catches and hacks (https://github.com/tensorpack/tensorpack/tree/master/examples/basics).

If you have a very large model which might be hard to migrate to tensorpack layers or `tf.layers`, chances are it may not work with the style of `mnist-keras.py` since you saw there are a lot of issues in the compatibility between the two libraries. My guess is that the `mnist-keras-v2` style may be less prone to such error.

see--:
The keras models are here: https://github.com/see--/keras-centernet. I now want to reproduce some training results. I thought about using 100% tensorpack, but I already defined the models. It would be a lot duplicate code. I can't use `tf.data` neither as it doesn't have all the `cv2` ops that are required.

> My guess is that the mnist-keras-v2 style may be less prone to such error.

I just started using tensorpack. It is easier to just follow the examples, and they mostly use the config + trainer style. This style is also better tested & documented. I actually thought that these were most of the rough edges :) Please let me know if you are aware of more.

PR #1283, opened 2019-07-26
Add extract.py

cocopambag:
Hi, ppwwyyxx!
 Recently, I usually use to tensorpack example of FasterRCNN.  tensorpack is a breakthrough. I like it.

 Anyway, Faster-RCNN is a popular object detection model. so, many people perhaps use this example.
Sometimes, who use this example want to extract graph. I did too. 

 Of course there are examples of [extraction](https://github.com/tensorpack/tensorpack/blob/master/examples/basics/export-model.py).
but, I wanted to make it easy for anyone to use. so, i create pull request.
Thank you!

### DIFF
--- a/examples/FasterRCNN/extract.py
+++ b/examples/FasterRCNN/extract.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+## export.py
+import argparse
+import tensorflow as tf
+
+from tensorpack import *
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Hi!

Thanks for your contribution! Since most the code are identical from `predict.py`, could you add a new flag in `predict.py` to simplify the implementation?
A flag like `./predict.py --export compact or --export serving` would be nice.

cocopambag:
You are right.
I recreate pull request!

PR #1288, opened 2019-07-26
Feat. extract graph

cocopambag:
Hi, ppwwyyxx!
Recently, I usually use to tensorpack example of FasterRCNN. tensorpack is a breakthrough. I like it.

Anyway, Faster-RCNN is a popular object detection model. so, many people perhaps use this example.
Sometimes, who use this example want to extract graph. I did too.

Of course there are examples of extraction.
but, I wanted to make it easy for anyone to use. so, i create pull request and modified predict.py
Thank you!


cocopambag:
well, Is this problem different OS?
when i processed tensorpack, i use ubuntu. but, when i support pull request, i use windows.

PR #1316, opened 2019-09-06
Save checkpoints and event files on Google Cloud Storage

dan-anghel:
Removed normalizing the path where the TensorFlow checkpoints, model and TensorBoard event files are saved in order to be able to save these to remote storages in the Cloud, like Google Cloud Storage, which are references with paths like gs://BUCKET_NAME/path/to/folder.


ppwwyyxx:
`normpath` is needed on windows.

Perhaps a solution is to add a custom `normpath` in `utils/fs.py`, which skips schema based paths (anything that contains `://`).

ppwwyyxx:
Looks great! Thanks for the fix!

dan-anghel:
Awesome, thank you for having merged this so quickly!

PR #1323, opened 2019-09-16
Transpose Convolution Valid/Full Padding

yg320:
Tensporpack implementation for Transpose Convolution is used when TF-version > 1.12.
In this implementation, output shape is calculated without taking padding into consideration.

I added support for valid/full padded Transpose Convolution in this pull request (which is consistent with tf.layers implementation).

### DIFF
--- a/tensorpack/models/conv2d.py
+++ b/tensorpack/models/conv2d.py
@@ -209,36 +209,34 @@ def Conv2DTranspose(
         strides2d = shape2d(strides)
         kernel_shape = shape2d(kernel_size)
 
-        assert padding.lower() in ['valid', 'full', 'same'], "Padding {} is not supported!".format(padding)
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
Thanks for catching this! I just added a unittest in f7a79d48cb00e9bda8d5f6928348d0a90788edc2 for such errors. Could you add "valid/full" to the test as well and make sure it passes?

yg320:
Done.
Also, I decided to remove support for 'full' padding since neither tf.layers.Conv2DTranspose nor tf.nn.conv2d_transpose seem to support that.

ppwwyyxx:
It looks like the commit "Removed NCHW test from TestConv2DTranspose" actually also removes the tests for valid padding.

yg320:
Sorry for that, I added it back

PR #1386, opened 2020-01-20
fix input dtype not matching variable dtype

gehuangyi20:
By default, all tensorflow variable has the dtype `tf.float32`. So, if the input has dtype `tf.float16`, then in the branch I edited, there is a dtype miss-matching between the input and the variable.

It will cause the error

> TypeError: Input 'filter' of 'Conv2D' Op has type float32 that does not match type float16 of argument 'input'.

The bug is discovered by running the code under repo [https://github.com/facebookresearch/ImageNet-Adversarial-Training](ImageNet-Adversarial-Training) with the following command

```bash
python3 main.py --data [imagenet training data dir] \
        --use-fp16xla --batch 256 --attack-iter 30 --attack-epsilon 16.0 --arch ResNeXtDenoiseAll -d 50
```

### DIFF
--- a/tensorpack/models/conv2d.py
+++ b/tensorpack/models/conv2d.py
@@ -241,7 +241,7 @@ def Conv2DTranspose(
                               filters]
 
         inputs_dtype = inputs.dtype
-        W = tf.get_variable('W', kernel_shape + [filters, channels_in], 
+        W = tf.get_variable('W', kernel_shape + [filters, channels_in],
                             dtype=inputs_dtype, initializer=kernel_initializer)
         if use_bias:
             b = tf.get_variable('b', [filters], dtype=inputs_dtype, initializer=bias_initializer)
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
Thanks! We've only tested regular ResNet and it looks like the group conv implementation & convtranspose does not work for fp16.

However your PR is not working yet: https://github.com/tensorpack/tensorpack/pull/1386/checks?check_run_id=399832218
Please make sure it passes flake8, and run the command in your issue again. In the log you're supposed to see https://github.com/facebookresearch/ImageNet-Adversarial-Training/blob/1ad68f08b8533083b0b8823ac3fd85cede191646/adv_model.py#L96 printed.

gehuangyi20:
I fixed the typo and code format so that all checks are passed. My code can use the fp16 right now.

Issue #1411, opened 2020-03-15
Train with Negative Dataset for Cascade R-CNN


mdv3101:
I am using tensorpack/examples/FasterRCNN for cascade RCNN training. I am training it for single class (say object). My dataset has few unlabelled images (negative dataset). I want to include them during training. Since they do not have any 'object' present in them, the bounding box has value [0,0,0,0]. Is setting label to 'BG' and label_id to 0 is enough?
Can you please tell the procedure?

ppwwyyxx:
If there are no objects you should leave "boxes/labels" empty rather than an array full of zeros.

ppwwyyxx:
https://github.com/tensorpack/tensorpack/blob/d89b6f074d3a37a88aa35df49da58f934bd694f8/examples/FasterRCNN/dataset/dataset.py#L20-L26

mdv3101:
Setting "boxes" to empty generates the following error (during loading of annotations)
`File "/home/madhav3101/cascade/tensorpack/examples/FasterRCNN/dataset/coco.py", line 149, in _add_detection_gt
    x1, y1, w, h = list(map(float, obj['bbox']))
ValueError: not enough values to unpack (expected 4, got 0)`

ppwwyyxx:
The above comment is for custom format.
If you're using COCO format, what you do is to simply not create an annotation for the image in the json file.

mdv3101:
If i am not getting wrong, For the coco format all i have to do is:
* In the 'images' of json, create an entry for every image in dataset (both positive and negative images). 
* In the 'annoatations' of json, create entry for every positive image (i.e. image which contains the bounding box)

If i follow these steps, will it going to include negative images (i.e. images which doesn't contain the object, only background) during the training. Or is it just going to ignore them and keeps on training using the positive images only?

ppwwyyxx:
The steps are correct and you'll also need to remove https://github.com/tensorpack/tensorpack/blob/d89b6f074d3a37a88aa35df49da58f934bd694f8/examples/FasterRCNN/data.py#L346-L348 otherwise those images will be filtered

mdv3101:
As you suggested, I have modified roidbs to include all the images. But now i am getting the following error.

  0%|          |0/1000[00:00<?,?it/s]2020-03-16 13:10:52.352836: I tensorflow/stream_executor/dso_loader.cc:152] successfully opened CUDA library libcublas.so.10.0 locally
[...]

PR #1424, opened 2020-04-15
Prevent filtering of empty images

mdv3101:
As per the issue #1411 , to enable negative dataset training, I have made few changes.

### DIFF
--- a/examples/FasterRCNN/data.py
+++ b/examples/FasterRCNN/data.py
@@ -345,7 +345,10 @@ def get_train_dataflow():
     # Filter out images that have no gt boxes, but this filter shall not be applied for testing.
     # The model does support training with empty images, but it is not useful for COCO.
     num = len(roidbs)
-    roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) >= 0, roidbs))
+    if cfg.DATA.FILTER_EMPTY_ANNOTATIONS:
+        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) >= 0, roidbs))
+    else:
+        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) > 0, roidbs))        
     logger.info(
         "Filtered {} images which contain no non-crowd groudtruth boxes. Total #images for training: {}".format(
             num - len(roidbs), len(roidbs)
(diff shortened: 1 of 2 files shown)


ppwwyyxx:
The original code is meant to filter out negative samples on purpose.

To support the other behavior, one way is to add a config option `DATA.FILTER_EMPTY_ANNOTATIONS`.

examples/FasterRCNN/data.py
```diff
@@ -345,7 +345,10 @@ def get_train_dataflow():
     # Filter out images that have no gt boxes, but this filter shall not be applied for testing.
     # The model does support training with empty images, but it is not useful for COCO.
     num = len(roidbs)
-    roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) > 0, roidbs))
+    if cfg.DATA.FILTER_EMPTY_ANNOTATIONS:
+        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) >= 0, roidbs))
+    else:
+        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) > 0, roidbs))
```
ppwwyyxx, line 351:
```suggestion
    if cfg.DATA.FILTER_EMPTY_ANNOTATIONS:
        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) > 0, roidbs))
```

`FILTER_EMPTY_ANNOTATIONS=False` means nothing need to be done

examples/FasterRCNN/data.py
```diff
@@ -345,7 +345,10 @@ def get_train_dataflow():
     # Filter out images that have no gt boxes, but this filter shall not be applied for testing.
     # The model does support training with empty images, but it is not useful for COCO.
     num = len(roidbs)
-    roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) > 0, roidbs))
+    if cfg.DATA.FILTER_EMPTY_ANNOTATIONS:
+        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) > 0, roidbs))
+    else:
+        roidbs = list(filter(lambda img: len(img["boxes"][img["is_crowd"] == 0]) >= 0, roidbs))
```
ppwwyyxx, line 351: the two lines can be deleted because `FILTER_EMPTY_ANNOTATIONS=False` means "do nothing".
`len()` is always >=0, so nothing will be filtered.

PR #1487, opened 2020-09-30
Improve compatibility to tensorflow 2.3

philippwerner:
When trying to run my tensorpack code (which I developed with tensorflow 1) with tensorflow 2.3, I got two errors inside tensorpack. I fixed both errors in these commits. Now my tensorpack code runs fine with tensorflow 2.3.

Best,
Philipp

PS: Thanks for developing tensorpack! It is a great library!


ppwwyyxx:
Thanks!

PR #1538, opened 2021-08-10
TensorFlow Version 1 Compatibility Fix

ajayvohra2005:
Added TensorFlow V1 compatibility fix to ```tensorpack/train/trainers.py``` and ```examples/FasterRCNN/eval.py``` so FasterRCNN model can be trained with latest version of TensorFlow 2.


ppwwyyxx:
Hi thanks that looks good. Could you make sure `flake8 .` checks can pass?

(there is also an unrelated unittest failure, please disregard)

